PR #7, opened 2024-02-22
provide some test utility functions

imrn99:
# Description

Add basic functions to generate sample meshes for future tests & benchmarks

Before merging:
- [x] Finish the `splitsquare_two_map` implementation: bcae200
- [x] Add tests to ensure generated map consistency
- [x] Add a figure in the `splitsquare_two_map` documentation
- [x] Resolve some merging issues probably

## Scope

- [x] Code: `honeycomb-core`, mostly in the `generation` module

## Type of change

- [x] New feature(s)

## Other

...

## Necessary follow-up

- [x] Needs documentation: update the member page in the UG

## Mentions

...


honeycomb-core/src/generation.rs
```diff
@@ -0,0 +1,441 @@
...
+        assert_eq!(cmap.beta::<2>(15), 0);
+        assert_eq!(cmap.beta::<2>(16), 10);
+    }
+
+    #[test]
+    fn splitsquare_two_map_correctness() {
+        let cmap: TwoMap<1> = splitsquare_two_map(2);
+
+        // hardcoded because using a generic loop & dim would just mean
+        // reusing the same pattern as the one used during construction
```
cedricchevalier19, line 295: And it is also easier to locate the error than when hidden in macros or loops.

honeycomb-core/src/twomap.rs
```diff
@@ -1255,7 +1327,41 @@ impl<const N_MARKS: usize> TwoMap<N_MARKS> {
...
+///
+/// Return the approximate size of the structure **in bytes**.
+///
+/// # Example
+///
+/// ```text
+///
+/// ```
+///
+pub fn map_size<const N_MARKS: usize>(map: &TwoMap<N_MARKS>) -> usize {
```
cedricchevalier19, line 1350: Is it a function instead of a method to handle other kind of maps in the future?

honeycomb-core/src/lib.rs
```diff
@@ -11,6 +11,7 @@
 
 pub mod dart;
 pub mod embed;
+pub mod generation;
```
cedricchevalier19, line 14: Not sure if it will stay there. Looks more an external utility to me.

cedricchevalier19 (approved):

honeycomb-core/src/lib.rs
imrn99: Probably, it could be in the lib associated to the crate containing benchmarks.

honeycomb-core/src/twomap.rs
imrn99: I initially thought this would be a utility function, but a detailed breakdown of memory usage will require reading into the private fields; I probably will turn it into a method when writing memory-related benchmarks

PR #15, opened 2024-03-05
feat: add custom 2D vec structure

imrn99:
# Description

Add a custom structure representing 2D coordinates: `Coords2`.  The type used by the map is still `Vertex2`, but it now stands as an alias to the new structure. This is done in case we ever discriminate vectors from vertices.

The structure uses `f64` as a floating type by default but it can be changed by enabling the `single_precision` feature for the core crate.

## Scope

- [x] Code: all features in `honeycomb-core`, other changes are just support/usage of the new structure
- [ ] Documentation

## Type of change

- [x] New feature(s)

## Other

- [x] New dependency

## Necessary follow-up

...

## Mentions

...

### DIFF
--- a/honeycomb-core/src/coords.rs
+++ b/honeycomb-core/src/coords.rs
@@ -31,7 +31,7 @@ cfg_if::cfg_if! {
 ///
 /// ```
 ///
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Default)]
 pub struct Coords2 {
     /// First coordinate
     pub x: FloatType,
--- a/honeycomb-core/src/coords.rs
+++ b/honeycomb-core/src/coords.rs
@@ -39,6 +39,28 @@ pub struct Coords2 {
     pub y: FloatType,
 }
 
+impl Coords2 {
+    pub fn norm(&self) -> FloatType {
+        (self.x * self.x + self.y * self.y).sqrt()
+    }
+
+    pub fn unit_dir(&self) -> Coords2 {
+        *self / self.norm()
+    }
+
+    pub fn dot(&self, other: &Coords2) -> FloatType {
+        self.x * other.x + self.y * other.y
+    }
+
+    // should this consume self & other?
+    pub fn cross(&self, other: &Coords2) -> Coords2 {
+        Self {
+            x: self.x * other.y - self.y * other.x,
+            y: other.x * self.y - other.y * self.x,
+        }
+    }
+}
+
 // Building traits
 
 impl From<(FloatType, FloatType)> for Coords2 {
--- a/honeycomb-core/src/coords.rs
+++ b/honeycomb-core/src/coords.rs
@@ -10,7 +10,13 @@
 
 // ------ CONTENT
 
-pub type FloatType = f64;
+pub type FloatType = cfg_if::cfg_if! {
+    if #[cfg(feature = "single_precision")] {
+        f32
+    } else {
+        f64
+    }
+};
 
 pub struct Coords2 {
     pub x: FloatType,
--- a/honeycomb-core/src/coords.rs
+++ b/honeycomb-core/src/coords.rs
@@ -31,7 +31,7 @@ cfg_if::cfg_if! {
 ///
 /// ```
 ///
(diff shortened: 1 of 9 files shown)


honeycomb-core/src/coords.rs
```diff
@@ -0,0 +1,254 @@
...
+/// assert_eq!(x_plus_y.unit_dir(), Coords2::from((1.0 / two.sqrt(), 1.0 / two.sqrt())));
+/// ```
+///
+#[derive(Debug, Clone, Copy, Default, PartialEq)]
+pub struct Coords2 {
+    /// First coordinate
+    pub x: FloatType,
+    /// Second coordinate
+    pub y: FloatType,
+}
```
cedricchevalier19, line 50: I would prefer a templated version of `Coords2` and not have a general switch and only one data type.

cedricchevalier19, line 50: Another (first) version is to stick to encapsulate an array of size 2.

cedricchevalier19 (changes requested):
Can we discuss about a templated version tomorrow?

honeycomb-core/src/embed.rs
```diff
@@ -95,7 +96,7 @@ pub enum UnsewPolicy {
 }
 
 /// Type definition for 2D vertex representation.
-pub type Vertex2 = [f64; 2];
+pub type Vertex2<T> = Coords2<T>;
```
cedricchevalier19, line 99: Good to have a separate type. You can later add attribute like a name or different coordinates systems.

honeycomb-core/src/twomap.rs
```diff
@@ -1117,14 +1111,8 @@ impl<const N_MARKS: usize> TwoMap<N_MARKS> {
...
-                        let new_lvertex = [
-                            (lvertex[0] + b1_rvertex[0]) / 2.0,
-                            (lvertex[1] + b1_rvertex[1]) / 2.0,
-                        ];
-                        let new_rvertex = [
-                            (rvertex[0] + b1_lvertex[0]) / 2.0,
-                            (rvertex[1] + b1_lvertex[1]) / 2.0,
-                        ];
+                        let new_lvertex = (lvertex + b1_rvertex) / T::from(2.0).unwrap();
+                        let new_rvertex = (rvertex + b1_lvertex) / T::from(2.0).unwrap();
```
cedricchevalier19, line 1115: Perhaps adding a mid_vertex function or something like this (if useful elsewhere).

cedricchevalier19 (approved):
Good addition!

honeycomb-core/src/coords.rs
```diff
@@ -0,0 +1,325 @@
...
+            y: -self.y,
+        }
+    }
+}
+
+impl<T: CoordsFloat> Sum<Coords2<T>> for Coords2<T> {
+    fn sum<I: Iterator<Item = Self>>(iter: I) -> Coords2<T> {
+        iter.fold(Self::default(), |c1, c2| c1 + c2)
+    }
+}
```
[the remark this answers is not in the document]
imrn99: This is useful to compute the center of a cell (face or volume) by directly summing elements of an iterator & dividing by its length. I'm not sure if it is useful for some meshing algorithm, but I will at least use it for rendering.

honeycomb-core/src/twomap.rs
imrn99: There are actually a lot of places where I'm doing this, I've added a function in 1fdda5b

honeycomb-core/src/coords.rs
```diff
@@ -0,0 +1,325 @@
...
+///
+#[derive(Debug, Clone, Copy, Default, PartialEq)]
+pub struct Coords2<T: CoordsFloat> {
+    /// First coordinate
+    pub x: T,
+    /// Second coordinate
+    pub y: T,
+}
+
+impl<T: CoordsFloat> Coords2<T> {
```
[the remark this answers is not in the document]
imrn99: I think we will make the distinction at some point, but this probably requires some breaking changes. I'll keep it for a later refactor.

PR #16, opened 2024-03-08
refactor: change structure used to track unused slots in the map

imrn99:
# Description

Change to the following attributes of `TwoMap`:

- `free_darts: Vec<DartIdentifier>`: rename to `unused_darts` & change type to `BTreeSet`
- `free_vertices: Vec<VertexIdentifier>`: rename to `unused_vertices` & change type to `BTreeSet`

New names are less ambiguous and sets are a better fit for our use-case.

This should probably be merged after #15.

## Scope

- [x] Code: `honeycomb-core`

## Type of change

- [x] Refactor

## Necessary follow-up

...

## Mentions

...

### DIFF
--- a/honeycomb-core/src/twomap.rs
+++ b/honeycomb-core/src/twomap.rs
@@ -154,7 +154,7 @@ const TWO_MAP_BETA: usize = 3;
 /// assert!(map.is_free(d6));
 ///
 /// // create the corresponding three vertices
-/// let v4 = map.add_vertex(Some([15.0, 0.0])); // v4
+/// let v4 = map.add_vertex(Some([15.0, 0.0].into())); // v4
 /// let v5 = map.add_vertices(2); // v5, v6
 /// let v6 = v5 + 1;
 /// map.set_vertex(v5, [5.0, 10.0]); // v5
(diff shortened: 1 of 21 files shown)


honeycomb-core/src/twomap.rs
```diff
@@ -294,15 +295,15 @@ impl<const N_MARKS: usize> TwoMap<N_MARKS> {
...
     ///
     pub fn new(n_darts: usize, n_vertices: usize) -> Self {
-        let vertices = vec![[0.0, 0.0]; n_vertices];
+        let vertices = vec![Vertex2::from([0.0, 0.0]); n_vertices];
         let betas = vec![[0; TWO_MAP_BETA]; n_darts + 1];
 
         Self {
             vertices,
-            free_vertices: Vec::with_capacity(n_vertices),
+            unused_vertices: BTreeSet::new(),
```
cedricchevalier19, line 303: You loose the pre-allocation here, is it important?

honeycomb-core/src/twomap.rs
```diff
@@ -794,7 +795,7 @@ impl<const N_MARKS: usize> TwoMap<N_MARKS> {
     /// See [TwoMap] example.
     ///
     pub fn remove_vertex(&mut self, vertex_id: VertexIdentifier) {
-        self.free_vertices.push(vertex_id);
+        assert!(self.unused_vertices.insert(vertex_id));
```
cedricchevalier19, line 798: Add a comment to explain the goal of this assert, or comment the failure in the documentation above.

cedricchevalier19 (approved):
Comment on the speed-up of this development.

honeycomb-core/src/twomap.rs
imrn99: Added doc, comment and some unit tests for this in 7fafa42.

honeycomb-core/src/twomap.rs
imrn99: I will run benchmarks and compare 0.1.1 to 7fafa42 (which I plan on making 0.1.2 with) . I will add relevant comparisons somewhere, not sure where yet.

PR #17, opened 2024-03-12
feat: implement a 2D combinatorial map visualizer

imrn99:
# Description

Add a new member to the workspace: `honeycomb-render`. This crate implements all that is needed to visualize the structure provided by `honeycomb-core`: `TwoMap`.

To-do:
- [x] finish the conversion code in the `honeycomb-render::handle` module
- [x] wire the map reference, to the runner, to the handle, to the render pipeline
- [x] expose a `RenderParameters` argument for the user to provide
- [x] add minimal usage instructions (for regular usage ~and web assembly target~)
- [x] add an example
- [x] fix the missing dart problem (cf. ac808cd)

## Scope

- [x] Code: `honeycomb-render`
- [x] Repository: new member

## Type of change

- [x] New feature(s)

## Other

- [x] New dependency

## Necessary follow-up

- [x] Documentation: Both for the crate & in the user guide / readme
- [x] Other: The `cgmath` dependency seems deprecated, it would be wise to look into it

## Mentions

...

### DIFF
--- a/honeycomb-render/src/camera.rs
+++ b/honeycomb-render/src/camera.rs
@@ -21,6 +21,14 @@ pub const OPENGL_TO_WGPU_MATRIX: cgmath::Matrix4<f32> = cgmath::Matrix4::new(
     0.0, 0.0, 0.0, 1.0,
 );
 
+cfg_if::cfg_if! {
+    if #[cfg(target_os = "linux")] {
+        const SPEED_FACTOR: f32 = 0.005;
+    } else {
+        const SPEED_FACTOR: f32 = 0.05;
+    }
+}
+
 pub struct Camera {
     pub eye: cgmath::Point3<f32>,
     pub target: cgmath::Point3<f32>,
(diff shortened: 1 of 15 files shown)


honeycomb-render/src/camera.rs
```diff
@@ -0,0 +1,180 @@
...
+    0.0, 1.0, 0.0, 0.0,
+    0.0, 0.0, 0.5, 0.5,
+    0.0, 0.0, 0.0, 1.0,
+);
+
+cfg_if::cfg_if! {
+    if #[cfg(target_os = "linux")] {
+        pub const SPEED_FACTOR: f32 = 0.005;
+    } else {
+        pub const SPEED_FACTOR: f32 = 0.05;
```
cedricchevalier19, line 28: Just being curious: why ?

cedricchevalier19 (approved):
Looks interesting, I trust you for the technical aspects.

honeycomb-render/src/camera.rs
imrn99: I tested the visualizer on my personal laptop (running MacOS) and there is a severe disparity in the camera speed at runtime (for the same value in the code). 

Aside from the OS difference, the keyboard sensitivity may also play a part, but I don't think it can explain that much of a difference.

PR #18, opened 2024-03-14
feat: implement an orbit structure for more efficient geometry computations

imrn99:
# Description

Implement a generic orbit structure that can be used for consistent, standardized cell computation. The algorithm used is detailed in the documentation.

The TwoMap implementation is left unchanged at the moment since replacing its code would probably lead to breaking changes. 

To do:
- [x] add more tests
- [x] add documentation

## Scope

- [x] Code: `honeycomb-core`

## Type of change

- [x] New feature(s)
- [x] Testing

## Other

...

## Necessary follow-up

- [x] Needs documentation
- [x] Other: Benchmark new methods

## Mentions

...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,9 +15,9 @@
   computation is done through the `Iterator` trait implementation.
 - `OrbitPolicy<'a>` - Enum used to specify the beta functions used by an orbit.
   It currently does not not support compositions.
-- New (temporary?) method for `TwoMap`: `beta_bis`. It works exactly like the
-  regular `beta` method except the beta index is passed as an argument instead
-  of a `const` generic.
+- New (temporary?) method for `TwoMap`: `beta_runtime`. It works by redirecting
+  to the original `beta` method, using match block and a beta identifier provided
+  at runtime.
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Change Log
 
+## To be released
+
+### Repository Changes
+
+...
+
+### New Features
+
+#### honeycomb-core
+
+- `Orbit<'a, const N_MARKS: usize, T: CoordsFloat>` - Generic implementation for
+  2D orbit computations. The structure itself only contains meta-data, the orbit
+  computation is done through the `Iterator` trait implementation.
+- `OrbitPolicy<'a>` - Enum used to specify the beta functions used by an orbit.
+  It currently does not not support compositions.
+- New (temporary?) method for `TwoMap`: `beta_bis`. It works exactly like the
+  regular `beta` method except the beta index is passed as an argument instead
+  of a `const` generic.
+
 ## 0.1.2
 
 ### Repository Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@
   regular `beta` method except the beta index is passed as an argument instead
   of a `const` generic.
 
+---
+
 ## 0.1.2
 
 ### Repository Changes
@@ -43,6 +45,8 @@
 #### honeycomb-utils
 
 - update content according to features introduced in #15, #16
+
+---
 
 ## 0.1.1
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,26 @@
...
+### New Features
+
+#### honeycomb-core
+
+- `Orbit<'a, const N_MARKS: usize, T: CoordsFloat>` - Generic implementation for
+  2D orbit computations. The structure itself only contains meta-data, the orbit
+  computation is done through the `Iterator` trait implementation.
+- `OrbitPolicy<'a>` - Enum used to specify the beta functions used by an orbit.
+  It currently does not not support compositions.
+- New (temporary?) method for `TwoMap`: `beta_bis`. It works exactly like the
```
cedricchevalier19, line 18: `beta_bis` name is not very good.
I would rather have `beta_rt` or `beta_runtime` and `beta_ct` (compile time?) to have an clearer idea of whats going on. Or `beta_generic` or `beta_dispatch`.

honeycomb-core/src/twomap.rs
```diff
@@ -102,9 +100,9 @@ const TWO_MAP_BETA: usize = 3;
 /// let (v1, v2, v3): (VertexIdentifier, VertexIdentifier, VertexIdentifier) = (0, 1, 2);
 ///
 /// // place the vertices in space
-/// map.set_vertex(v1, [0.0, 0.0]);
-/// map.set_vertex(v2, [0.0, 10.0]);
-/// map.set_vertex(v3, [10.0, 0.0]);
+/// map.set_vertex(v1, [0.0, 0.0]).unwrap();
+/// map.set_vertex(v2, [0.0, 10.0]).unwrap();
+/// map.set_vertex(v3, [10.0, 0.0]).unwrap();
```
cedricchevalier19, line 105: Future work: proper error handling, with `?` for example.

honeycomb-core/src/twomap.rs
```diff
@@ -379,6 +383,32 @@ impl<const N_MARKS: usize, T: CoordsFloat> TwoMap<N_MARKS, T> {
...
+    ///
+    /// The method will panic if *i* is not 0, 1 or 2.
+    ///
+    /// # Example
+    ///
+    /// See [TwoMap] example.
+    ///
+    pub fn beta_bis(&self, i: u8, dart_id: DartIdentifier) -> DartIdentifier {
+        assert!(i < 3);
+        self.betas[dart_id as usize][i as usize]
```
cedricchevalier19, line 409: Ok, `betas` is simple enough but a `match` dispatcher that forward to the corresponding compile time beta function could also be considered.

honeycomb-core/src/orbits.rs
```diff
@@ -0,0 +1,267 @@
...
+pub enum OrbitPolicy<'a> {
+    /// 0-cell orbit.
+    Vertex,
+    /// 1-cell orbit.
+    Edge,
+    /// 2-cell orbit.
+    Face,
+    /// User-defined orbit. The integers correspond to the indices of the
+    /// beta functions that should be used for the computation.
+    Custom(&'a [u8]),
```
cedricchevalier19, line 27:
```suggestion
    /// Ordered array of beta functions that define the orbit
    Custom(&'a [u8]),
```

Not sure it is mathematically correct, but I think the order thing is important.

cedricchevalier19 (approved):
Not totally sure about the naming.

honeycomb-core/src/orbits.rs
imrn99: This is actually a much better description!

CHANGELOG.md
imrn99: I renamed `beta_bis` to `beta_runtime` and cahnged its implementation to use a match block. I don't wan't to rename the original beta function at the moment though.

honeycomb-core/src/twomap.rs
imrn99: I tried this initially, but it seems the compiler isn't fond of having a returned value in doc tests. I'm open to suggestion on how to resolve this

cedricchevalier19: You can look at Métis-rs or coupe for examples.

PR #24, opened 2024-03-18
refactor: remove builtin marks

imrn99:
# Description

Remove the marks built-in the `DartData` structure. This results in many function / structure signature changes, most notably:
- `Cmap2`
- `Orbit2`
- Both `honeycomb-utils::generation` functions
- `CMap2RenderHandle` in `honeycomb-render`

## Scope

- [x] Code
- [x] Documentation

## Type of change

- [x] Refactor

## Other

- [x] Breaking change

## Necessary follow-up

...

## Mentions

...

### DIFF
--- a/honeycomb-core/src/twomap.rs
+++ b/honeycomb-core/src/twomap.rs
@@ -13,7 +13,8 @@
 // ------ IMPORTS
 
 use std::collections::BTreeSet;
-use std::{fs::File, io::Write, sync::atomic::AtomicBool};
+#[cfg(feature = "benchmarking_utils")]
+use std::{fs::File, io::Write};
 
 use crate::coords::CoordsFloat;
 use crate::{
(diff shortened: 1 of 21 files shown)


honeycomb-core/src/twomap.rs
```diff
@@ -1289,14 +1288,15 @@ impl<const N_MARKS: usize, T: CoordsFloat> CMap2<N_MARKS, T> {
     ///
     pub fn build_all_faces(&mut self) -> usize {
         self.faces.clear();
+        let mut marked = BTreeSet::<DartIdentifier>::new();
```
cedricchevalier19, line 1291: So this is the place where marks where used?

cedricchevalier19 (approved):
I am glad that you removed this parameter.

honeycomb-core/src/twomap.rs
imrn99: Yes, in the reference implementation / algorithm, marks also were used for orbits but it doesn't really makes sense for a practical Rust implementation IMO.

PR #25, opened 2024-03-19
feat: add better spatial representation structures

imrn99:
# Description

Implement two new structures, both based on the `Coords2` struct: 

- `Vertex2` - a 2D vertex type
- `Vector2` - a 2D vector type

By defining those as wrappers around a `Coords2` value (instead of a simple type alias), we can introduce notions of homogeneity (e.g. `Vertex + Vector = Vertex`, `Vertex - Vertex = Vector`).

## Scope

- [x] Code: `honeycomb-core`

## Type of change

- [x] New feature(s)
- [x] Testing

## Other

...

## Necessary follow-up

- [x] Needs documentation: user guide & Vertex2 example
- [x] Other: types and structures are only implemented, not integrated to the `CMap2` structure; They should be integrated with another PR to avoid *mergeability* issues.

## Mentions

...


cedricchevalier19 (commented):
I think it is a good idea for robustness.

imrn99:
I rewrote `Vector2` methods to make them independent from the original `Coords2` methods (e1d3034). The latter are now tagged with a `#[cfg(test)]` & the documentation clearly states that the type should not be used directly (8a00ff4).

cedricchevalier19:
> I rewrote `Vector2` methods to make them independent from the original `Coords2` methods ([e1d3034](https://github.com/LIHPC-Computational-Geometry/honeycomb/commit/e1d3034ca7e65af88a7b05bfe5593113f7552086)).
> The latter are now tagged with a `#[cfg(test)]` & the documentation clearly states that the type should not be used directly ([8a00ff4](https://github.com/LIHPC-Computational-Geometry/honeycomb/commit/8a00ff4fa34f096f815be750f49d98a5f4616765)).

Why keeping them?

imrn99:
> > I rewrote `Vector2` methods to make them independent from the original `Coords2` methods ([e1d3034](https://github.com/LIHPC-Computational-Geometry/honeycomb/commit/e1d3034ca7e65af88a7b05bfe5593113f7552086)).
> > The latter are now tagged with a `#[cfg(test)]` & the documentation clearly states that the type should not be used directly ([8a00ff4](https://github.com/LIHPC-Computational-Geometry/honeycomb/commit/8a00ff4fa34f096f815be750f49d98a5f4616765)).
> 
> Why keeping them?

I tried to think of a justification to keep those methods in, but could not find any. I removed them & only kept the `std::ops` implementations as they are useful for `Vector2`, `Vertex2` & the rendering code.

PR #27, opened 2024-03-28
refactor: restructure core crate source code

imrn99:
# Description

Reorganize the structure of the core crate into thematically consistent modules. This does not affect the public API, only inner paths.

## Scope

- [x] Code:  `honeycomb`

## Type of change

- [x] Refactor

## Other

...

## Necessary follow-up

- [x] ~Other: Mention in changelog file~ done

## Mentions

...


honeycomb-core/src/cells/identifiers.rs
```diff
@@ -0,0 +1,42 @@
+//! Module short description
```
cedricchevalier19, line 1: Not sure about having a separate file for that.
These ids are mostly used as internal indices, so keeping them close to data structures also make sense.

cedricchevalier19 (approved):

honeycomb-core/src/cells/identifiers.rs
imrn99: I have grouped them here at the moment since I will be re-implementing data structures for attributes sooner than expected (see #30)

Issue #30, opened 2024-04-03
Enh: introduce a generic attribute system to associate data to i-cells


imrn99:
# Description

For future development, a generic attribute system should be coded. This will result in two things:
- a consistent internal implementation that allow us to write a more clear, concise implementation of sewing operations
- the possibility to expose an API to the user, which can then be used to add user-defined attribute to cells of the map

The single existing attribute (2D coordinates) can be implemented using that system to ensure its viability.

## Current progress

- [x] code generic traits that attributes will implement: #33 
- [x] code a storage structure defined for generic attributes: #34 
- [x] replace the current vertex storage by the new system: 
    - #35 
    - #36 
- Complete the changelog:
    - [x] PR #33 
    - [x] PR #34 
    - [x] PR #36 

## Information

- [x] Suggestion for a new feature

### Current state

As of now (4c0d5d2), the only attribute associated to cells are a set of 2D coordinates (associated to 0-cells/vertices). Those are hardcoded into the `CMap2` structure. The goal is to write the new system & use it to implement coordinates handling.

### Proposal

The system is made up of 3 components:

#### Attribute trait

~We can define a trait `Tr` (name TBD) that all attributes must implement in order to handle automatically value changes, e.g.~

Following a suggestion made by @cedricchevalier19, we can define two traits in order to separate the operational nature from the support entity of a given attribute. In practice, it would look like this:

```rust
pub trait AttributeLogic: Sized {
    /// merge two values of the attribute into a single one
    fn merge(lhs: Self, rhs: Self) -> Self;

    /// create two values of the attribute from a single one 
    fn split(lhs: Self) -> (Self, Self);

    /// default implementation in case the attributes 
    /// are not fully defined on relevant cells
    /// this can be overwritten by the user if needed
    fn merge_incomplete(lhs: Option<Self>) -> Self {
        lhs.unwrap_or(todo!())
    }
}

pub trait AttributeSupport: Sized {
    fn binds_to(&self) -> Cell;
}

pub enum Cell {
    Vertex,
    Edge,
    Face,
}
```

Note that the `Cell` enum is just a placeholder for something more relevant.

#### Attribute storage structure

This specific part of the structure is most likely where benchmarking and optimization will happen. Considering this, and the fact that custom attribute support will most likely require a builder struct (e.g. `CMap2Builder`), the first storage structure will be kept simple & its usage will be hardcoded into the current `CMap2` constructor. For now, the inner storage will most likely look like this

[...]

PR #33, opened 2024-04-09
feat: implement generic attribute traits

imrn99:
# Description

Add two traits: ~`AttributeLogic`~ `AttributeUpdate` and ~`AttributeSupport`~ `AttributeBind`. These will be used in order to introduce attribute genericity in the map implementation. 

## Scope

- [x] Code: `honeycomb-core`
- [x] Documentation

## Type of change

- [x] New feature(s)

## Other

...

## Necessary follow-up

- [x] Other: Refer to #30 

## Mentions

...

### DIFF
--- a/honeycomb-core/src/cells/attributes.rs
+++ b/honeycomb-core/src/cells/attributes.rs
@@ -6,6 +6,16 @@
 
 // ------ CONTENT
 
+pub trait AttributeLogic: Sized {
+    fn merge(lhs: Self, rhs: Self) -> Self;
+
+    fn split(lhs: Self) -> (Self, Self);
+
+    fn merge_undefined(lhs: Option<Self>) -> Self {
+        lhs.unwrap() // todo: choose a policy for default behavior
+    }
+}
+
 // ------ TESTS
 
 #[cfg(test)]
--- a/honeycomb-core/src/cells/attributes.rs
+++ b/honeycomb-core/src/cells/attributes.rs
@@ -4,6 +4,8 @@
 
 // ------ IMPORTS
 
+use crate::OrbitPolicy;
+
 // ------ CONTENT
 
 pub trait AttributeLogic: Sized {
@@ -14,6 +16,10 @@ pub trait AttributeLogic: Sized {
     fn merge_undefined(lhs: Option<Self>) -> Self {
         lhs.unwrap() // todo: choose a policy for default behavior
     }
+}
+
+pub trait AttributeSupport: Sized {
+    fn binds_to(&self) -> OrbitPolicy;
 }
 
 // ------ TESTS
--- a/honeycomb-core/src/cells/attributes.rs
+++ b/honeycomb-core/src/cells/attributes.rs
@@ -28,25 +28,25 @@ use crate::OrbitPolicy;
 /// }
 ///
 /// impl AttributeLogic for Temperature {
-///     fn merge(lhs: Self, rhs: Self) -> Self {
-///         Temperature { val: (lhs.val + rhs.val) / 2.0 }
+///     fn merge(attr1: Self, attr2: Self) -> Self {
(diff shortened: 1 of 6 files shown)


honeycomb-core/src/cells/attributes.rs
```diff
@@ -0,0 +1,107 @@
...
+///     fn split(lhs: Self) -> (Self, Self) {
+///         (lhs, lhs)
+///     }
+///
+///     fn merge_undefined(lhs: Option<Self>) -> Self {
+///         lhs.unwrap_or(Temperature { val: 0.0 })
+///     }
+/// }
+/// ```
+pub trait AttributeLogic: Sized {
```
cedricchevalier19, line 44: `AttributeLogic` is not very appealing. Can something like `AttributeMergeable` or `AttributeMergeTrait` or `AttributeMerge` be easier to understand.

So next question, is there a reason to tie `merge` and `split` together or can we have algorithm that can deal with only `merge` or only `split`?

honeycomb-core/src/cells/attributes.rs
```diff
@@ -0,0 +1,107 @@
...
+///         OrbitPolicy::Face
+///     }
+/// }
+/// ```
+pub trait AttributeSupport: Sized {
+    /// Return an [OrbitPolicy] that can be used to identify the kind of topological entity to
+    /// which the attribute is associated.
+    fn binds_to(&self) -> OrbitPolicy;
+}
+
```
cedricchevalier19, line 96: Perhaps we can improve coherency: `support` or `bind`, between the trait's name and its method.

honeycomb-core/src/cells/attributes.rs
```diff
@@ -0,0 +1,107 @@
...
+    ///
+    /// This trait and its methods were designed with the (un)sewing operation in mind. Their
+    /// purpose is to simplify the code needed to propagate updates of attributes affected by the
+    /// (un)sewing operation. Considering this context, as well as the definition of (un)linking
+    /// operations, this panic seems reasonable: If the darts you are sewing have totally undefined
+    /// attributes, you should most likely be linking them instead of sewing.
+    fn merge_undefined(lhs: Option<Self>) -> Self {
+        lhs.unwrap()
+    }
+}
```
cedricchevalier19, line 65: Comparing to `bind_to`, these are functions and not methods, is there any reason to treat them differently?

Also it is not very clear what `merge_undefined` can do with 0 or 1 attribute to *merge*.

cedricchevalier19 (approved):
I approve the technical changes but I am kind of annoyed by the trait names.

Rust is not very consistent when naming the traits (https://github.com/rust-lang/api-guidelines/discussions/28#) , but perhaps we still can improve that here.

honeycomb-core/src/cells/attributes.rs
imrn99: Agreed, I renamed the trait to `AttributeBind`, I'll try to think of an imperative for the new name of the other trait too.

honeycomb-core/src/cells/attributes.rs
imrn99: From what I understood from my discussion with @franck-ledoux, `merge` and `split` are always defined for a given attribute. Also, from a user API perspective, I think keeping the number of trait to a minimum is the best choice for both clarity & simplicity.

honeycomb-core/src/cells/attributes.rs
imrn99: I'm actually not sure why I made `binds_to` a method instead of a function. Both traits are meant to describe the **behavior of the attribute** (i.e. a type, structure, enum...), **not the behavior of a given instance of the attribute**. 

Defining these functions as methods also introduce some asymmetry to arguments which I don't think is coherent with what I described above.

PR #35, opened 2024-04-12
refactor: reimplement vertex in `CMap2` as attributes

imrn99:
# Description

Replace the current vertex storage (`Vec<Vertex2`) by the structure implemented in #34. The implementation is currently 
hardcoded into `CMap2` but the user should eventually be able to add its own attributes.

## Scope

- [x] Code: `honeycomb-core`

## Type of change

- [x] Refactor

## Other

...

## Necessary follow-up

...

## Mentions

...

### DIFF
--- a/honeycomb-core/src/cells/attribute_collections.rs
+++ b/honeycomb-core/src/cells/attribute_collections.rs
@@ -25,6 +25,7 @@ use num::ToPrimitive;
 ///
 /// todo
 ///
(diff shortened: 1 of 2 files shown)


imrn99:
After doing some work on the structure, I realized it would be much easier to directly rewrite the `CMap2` methods using the new attribute system AND the new index system. I'm closing this and keeping the branch as reference for the final implementation

PR #54, opened 2024-04-22
refactor: update visualization code to properly render faces

### LINKED ISSUE #22: Enh: display faces properly instead of using implicit coloring
# Description

## Information

- [x] Suggestion for a new feature

### Current state

The current implementation relies on a hard-coded color table in the shader, which is accessed using face IDs in order to give different colors to different faces. This is arbitrary, and the result isn't very pleasing

### Proposal

Use Rust's strong-typed enums to render darts to a given color, faces to another given color, etc. Colors can be attributed in the shader, using a system similar to the current one.

imrn99:
Before: 

Capture d’écran 2024-04-22 à 18 34 40

After:

Capture d’écran 2024-04-22 à 18 33 55

## Scope

- [x] Code: `honeycomb-render`

## Type of change

- [x] Fix: closes #22 
- [x] Refactor


honeycomb-render/src/handle.rs
```diff
@@ -56,108 +53,103 @@ impl<'a, T: CoordsFloat> CMap2RenderHandle<'a, T> {
...
+        self.dart_construction_buffer.extend(tmp);
     }
 
     #[allow(dead_code)]
     pub fn build_betas(&mut self) {
         todo!()
     }
 
+    pub fn build_faces(&mut self) {
+        // because there's no trianglefan priitive in the webgpu standard,
```
cedricchevalier19, line 126:
```suggestion
        // because there's no "trianglefan" primitive in the webgpu standard,
```

honeycomb-render/src/representations/intermediates.rs
```diff
@@ -0,0 +1,31 @@
...
+impl<T: CoordsFloat> IntermediateFace<T> {
+    pub fn new<I: Iterator<Item = Vertex2<T>>>(it: I) -> Self {
+        let tmp: Vec<Vertex2<T>> = it.collect();
+        let n_vertices = tmp.len();
+        let center = Vertex2::from(
+            tmp.iter()
+                .map(|vertex2: &Vertex2<T>| vertex2.into_inner())
+                .reduce(|coords1, coords2| coords1 + coords2)
+                .unwrap()
+                / T::from(n_vertices).unwrap(),
```
cedricchevalier19, line 23: `from` can fail or is it the division?
You can early exit if `n_vertices` is 0.

cedricchevalier19 (approved):

honeycomb-render/src/representations/intermediates.rs
imrn99: It's the `from` call that can fail. I still added a `match` block in case the face is made up of only 0 or 1 vertex. These case will naturally yield an empty iterator when attempting to render later in the execution flow.

PR #60, opened 2024-04-24
feat: add control over grid generation using a builder struct

imrn99:
- implement a builder structure `GridBuilder` to generate maps representing custom grids
- mark previous generation methods as deprecated and redirect the user to an alternative

## Scope

- [x] Code: `honeycomb-core` for the implementation, other members to remove deprecations

## Type of change

- [x] New feature(s)
- [x] Refactor

### DIFF
--- a/honeycomb-core/src/utils/generation.rs
+++ b/honeycomb-core/src/utils/generation.rs
@@ -596,6 +596,22 @@ mod tests {
     }
 
     #[test]
+    fn build_incomplete() {
+        assert!(GridBuilder::default()
+            .len_per_cell([1.0_f64, 1.0_f64, 1.0_f64])
+            .build2()
+            .is_err());
+        assert!(<GridBuilder<f64>>::default()
+            .n_cells([4, 4, 0])
+            .build2()
+            .is_err());
+        assert!(GridBuilder::default()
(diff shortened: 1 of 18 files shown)


dssgabriel (approved):

honeycomb-core/src/utils/generation.rs
```diff
@@ -433,7 +409,9 @@ impl<T: CoordsFloat> GridBuilder<T> {
                 )
             }
             (_, _, _) => {
-                return Err(BuilderError::MissingParameters("Insufficient building parameters, please specify two out of three grid parameters"));
+                return Err(BuilderError::MissingParameters(
+                    "GridBuilder: insufficient building parameters",
```
dssgabriel, line 413: Maybe a dumb question, but why use `'a` lifetime and not `'static` on your `BuilderError` enum? Your error string looks to be compile-time known.   
Future-proofing?

dssgabriel (approved):

honeycomb-core/src/utils/generation.rs
imrn99: This is a correct observation, I don't really need the runtime lifetime for my errors atm.

PR #63, opened 2024-04-26
chore: update changelog, bump versions

imrn99:
misc updates before release `0.2.1`

## Scope

- [x] Code: small changes to `honeycomb-examples` to match some deps update
- [x] Repository

## Type of change

- [x] Chore

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ members = [
 [workspace.package]
 edition = "2021"
 license = "MIT OR Apache-2.0"
-version = "0.2.0"
+version = "0.2.1"
 homepage = "https://lihpc-computational-geometry.github.io/honeycomb/"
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -36,7 +36,7 @@ num = "0.4.2"
 # benchmarks
 criterion = "0.5.1"
 iai-callgrind = "0.10.2"
-rand = "0.8.5"
+rand = "0.9.0-alpha.1"
```
dssgabriel, line 39: Why the need for `alpha`?

dssgabriel (approved):

Cargo.toml
imrn99: The `0.9.0` of the rand crate is not out yet and their devs seem to have released an alpha for people to use. Now was a good time to upgrade the dependency so I decided to do it and process a first batch a breaking changes that will be in `0.9.0`

PR #73, opened 2024-05-15
feat: implement basic vtk mesh reading 

imrn99:
add a new feature to the core crate: `io`. When enabled, the function `CMap2::from_vtk_file` is made available. It allows the user to initialize a combinatorial map from a VTK legacy file. 

There are a number of limitation on the input file, notable requirements include coherent cell orientation and exclusive cell types.

There are two I/O examples that can be used to verify the output map's validity

## Scope

- [x] Code: `honeycomb-core`, `honeycomb-examples`

## Type of change

- [x] New feature(s)

## Other

- [x] New dependency: `vtkio`


Cargo.toml
```diff
@@ -33,6 +33,7 @@ rustversion = "1.0.15"
 
 # core
 num = "0.4.2"
+vtkio = { git = "https://github.com/elrnv/vtkio.git" }
```
cedricchevalier19, line 36: Any reason to not target a released version?

cedricchevalier19 (approved):

Cargo.toml
imrn99: While their README indicates that the latest release is `0.7.0`, the latest found on crates.io is `0.6.3`. I didn't want to check how much changes there are between versions, so I just used the latest commit.

PR #74, opened 2024-05-15
refactor: remove `FloatType` and all of its usage

imrn99:
remove all `FloatType` usages in order to prepare for feature removal; also rewrote some tests using macros to still use both f32/f64 types.

## Scope

- [x] Code: `honeycomb-core`

## Type of change

- [x] Testing
- [x] Refactor

## Necessary follow-up

- [x] Other: remove the `_single_precision` feature and update CIs

### DIFF
--- a/honeycomb-core/src/spatial_repr/coords.rs
+++ b/honeycomb-core/src/spatial_repr/coords.rs
@@ -215,28 +215,40 @@ impl<T: CoordsFloat> IndexMut<usize> for Coords2<T> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::common::FloatType;
 
-    fn almost_equal(lhs: &Coords2<FloatType>, rhs: &Coords2<FloatType>) -> bool {
-        const EPS: FloatType = 10.0e-12;
-        ((lhs.x - rhs.x).abs() < EPS) & ((lhs.y - rhs.y).abs() < EPS)
+    macro_rules! almost_equal {
+        ($f1: expr, $f2: expr, $t: ty) => {
+            ($f1 - $f2).abs() < <$t>::EPSILON
+        };
+    }
+    macro_rules! almost_equal_vec {
+        ($lhs: expr, $rhs: expr, $t: ty) => {
+            almost_equal!($lhs.x, $rhs.x, $t) & almost_equal!($lhs.y, $rhs.y, $t)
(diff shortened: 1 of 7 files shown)


honeycomb-core/src/spatial_repr/coords.rs
```diff
@@ -215,28 +215,40 @@ impl<T: CoordsFloat> IndexMut<usize> for Coords2<T> {
...
-        let ref_value: Coords2<FloatType> = Coords2::from((3.0, 3.0));
-        assert!(almost_equal(&owned_sum, &ref_value));
-        assert!(almost_equal(&borrowed_sum, &ref_value));
-    }
+
+    macro_rules! almost_equal {
+        ($f1: expr, $f2: expr, $t: ty) => {
+            ($f1 - $f2).abs() < <$t>::EPSILON
+        };
+    }
```
cedricchevalier19, line 223: Not related to this PR, but it is not very correct.

Floating point numbers are designed to work with relative error, not absolute. So if you deal with large or small numbers, if comparison does not mean the same thing.

honeycomb-core/src/spatial_repr/vector.rs
```diff
@@ -291,77 +291,101 @@ impl<T: CoordsFloat> std::ops::Neg for Vector2<T> {
...
 mod tests {
     use super::*;
-    use crate::common::FloatType;
 
+    // utils
     macro_rules! almost_equal {
-        ($f1: expr, $f2: expr) => {
-            ((($f1 - $f2) as FloatType).abs() < FloatType::EPSILON)
+        ($f1: expr, $f2: expr, $t: ty) => {
+            ($f1 - $f2).abs() < <$t>::EPSILON
```
cedricchevalier19, line 298: Same thing about relative vs absolute comparison of floating point numbers.

cedricchevalier19 (approved):
I think it is a good idea to limit the number of features.

honeycomb-core/src/spatial_repr/coords.rs
imrn99: I found this [reference](https://floating-point-gui.de/errors/comparison/) for comparing floating point numbers. If it looks good to you, I'll implement something similar later since I plan to group all tests of the `spatial_repr` module.

PR #83, opened 2024-05-22
feat(honeycomb-core): add `CMap2::split_edge` for future algorithm implementations

imrn99:
Implement a new method, `CMap2::split_edge`, that split an existing edges into two, giving optional control over where the new intermediary vertex is placed.

## Scope

- [x] Code: `honeycomb-core`

## Type of change

- [x] New feature(s)

### DIFF
--- a/honeycomb-core/src/cmap2/advanced_ops.rs
+++ b/honeycomb-core/src/cmap2/advanced_ops.rs
@@ -5,7 +5,7 @@
 
 // ------ IMPORTS
 
-use crate::{CMap2, CoordsFloat, DartIdentifier, EdgeIdentifier, Vertex2};
+use crate::{CMap2, CoordsFloat, DartIdentifier, EdgeIdentifier, Vertex2, NULL_DART_ID};
 
 // ------ CONTENT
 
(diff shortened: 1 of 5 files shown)


honeycomb-core/src/cmap2/advanced_ops.rs
```diff
@@ -0,0 +1,87 @@
...
+impl<T: CoordsFloat> CMap2<T> {
+    /// Split an edge into to segments.
+    ///
+    /// <div class="warning">
+    /// This implementation is 2D specific.
+    /// </div>
+    ///
+    /// This method takes all darts of an edge and rebuild connections in order to create a new
+    /// point on this edge. The position of the point default to the midway point, but it can
+    /// optionally be specified.
```
cedricchevalier19, line 21: How do you ensure that the intermediate point is on the edge. Looks like a scalar parameter like 0.5 might be better to compute a point that is really located on the edge.

honeycomb-core/src/cmap2/advanced_ops.rs
```diff
@@ -0,0 +1,87 @@
...
+    ///
+    /// In order to minimize editing of embedded data, the original darts are kept to their
+    /// original vertices while the new darts are used to model the new point.
```
cedricchevalier19, line 24: Not very clear.

honeycomb-core/src/cmap2/advanced_ops.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// # Panics
+    ///
+    /// This method may panic if the edge upon which the operation is performed does not have two
+    /// defined vertices.a
+    pub fn split_edge(&mut self, edge_id: EdgeIdentifier, midpoint_vertex: Option<Vertex2<T>>) {
+        let d1 = edge_id as DartIdentifier;
+        let d2 = self.beta::<2>(d1);
+        // (*): unwrapping is ok because you probably shouldn't be using this method
+        //      if both vertices of the edge aren't defined
+        if d2 == NULL_DART_ID {
```
cedricchevalier19, line 35: Not clear: what does it means ?
variable names can be better than d1 and d2.

cedricchevalier19 (commented):
Change the title of the PR to include the name of the new method.

honeycomb-core/src/cmap2/advanced_ops.rs
imrn99: I thought that we might want to place vertices that are slightly offset from the original edges, but it's probably best to go with your idea and move vertices after if necessary.

honeycomb-core/src/cmap2/advanced_ops.rs
imrn99: I added an example that illustrates both of my points in d81ad37

honeycomb-core/src/cmap2/advanced_ops.rs
imrn99: I tried to clarify the comments (e471d51), but I'm not sure about how to name both of these darts

cedricchevalier19 (approved):

PR #88, opened 2024-05-30
refactor(core): deprecate old constructors

imrn99:
Mark previous constructors as deprecated since `CMapBuilder` covers all of their features.

## Scope

## Type of change

- [x] Refactor

### DIFF
--- a/honeycomb-core/src/cmapbuilder/grid/building_routines.rs
+++ b/honeycomb-core/src/cmapbuilder/grid/building_routines.rs
@@ -11,6 +11,7 @@ pub fn build_2d_grid<T: CoordsFloat>(
     [n_square_x, n_square_y]: [usize; 2],
     [len_per_x, len_per_y]: [T; 2],
 ) -> CMap2<T> {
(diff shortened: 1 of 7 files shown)


honeycomb-core/src/cmapbuilder/grid/building_routines.rs
```diff
@@ -11,6 +11,7 @@ pub fn build_2d_grid<T: CoordsFloat>(
     [n_square_x, n_square_y]: [usize; 2],
     [len_per_x, len_per_y]: [T; 2],
 ) -> CMap2<T> {
+    #[allow(deprecated)] // allow because CMap2::new() will still be available inside the crate
```
cedricchevalier19, line 14: Why do you not call another nonpublic new constructor (to which the old `new()` aliases) ?
This way, you can get rid of all these `allow(deprecated).`

imrn99: I didn't think of making the change then, I fixed this is #90

PR #89, opened 2024-05-31
feat!(core): implement `AttrStorageManager` struct and `AttributeStorage` trait 

imrn99:
- add a new structure: `AttrStorageManager`
- add a new trait: `AttributeStorage`; it describes a unified API for all collections used to store attributes
- add a new associated type to `AttributeBind`: `StorageType: AttributeStorage`

## Scope

- [x] Code: `honeycomb-core`

## Type of change

- [x] New feature(s)

## Other

- [x] Breaking change: `AttributeBind` has a new associated type which is used in the `AttrStorageManager` implementation

## Necessary follow-up

- [x] Needs testing
- [x] Other: 
    - accessing and editing single attributes is OK, but I haven't managed to implement a generic iteration over the attributes
    - Should the attribute traits be fused or some restriction be lifted? All the implementations are starting to pile up
    - Regular methods of the collections struct should be replaced by a direct implementation of `AttributeStorage` trait


honeycomb-core/src/common.rs
```diff
@@ -32,8 +33,11 @@ pub enum CMapError {
 
 /// Common trait implemented by types used for coordinate representation.
 pub trait CoordsFloat:
-    num::Float + Default + AddAssign + SubAssign + MulAssign + DivAssign
+    num::Float + Default + AddAssign + SubAssign + MulAssign + DivAssign + Debug + 'static
```
cedricchevalier19, line 36: Add a comment on why `'static` (for `TypeId`)

honeycomb-core/src/attributes/manager.rs
```diff
@@ -0,0 +1,134 @@
...
+            OrbitPolicy::Custom(_) => $slf.others.get_mut(&TypeId::of::<A>()),
+        };
+        let $id = probably_storage
+            .expect("E: could not find storage associated to the specified attribute type")
+            .downcast_mut::<<A as AttributeBind>::StorageType>()
+            .expect("E: could not downcast generic storage to specified attribute type");
+    };
+}
+
+#[allow(unused, missing_docs)]
```
cedricchevalier19, line 57: Add a small explanation of how it works.

cedricchevalier19 (approved):
Just add a few comments on the design.

PR #91, opened 2024-06-03
refactor!(core): change internal Vtk building routine to return a `Result`

imrn99:
- `build_2d_from_vtk` now return a `Result<CMap2, BuilderError>`
- a number the function's assertions are replaced with early `Err` returns
- `BuilderError` has been expanded with more (specialized) variants

## Scope

- [x] Code: `honeycomb-core`

## Type of change

- [x] Refactor

## Other

- [x] Breaking change: because of new `BuilderError` variants (& renamed old ones)

### DIFF
--- a/honeycomb-core/src/cmapbuilder/io/mod.rs
+++ b/honeycomb-core/src/cmapbuilder/io/mod.rs
@@ -6,7 +6,9 @@
 //! are supported, because of orientation and dimension restriction.
 
 // ------ IMPORTS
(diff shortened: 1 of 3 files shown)


honeycomb-core/src/cmapbuilder/io/mod.rs
```diff
@@ -52,11 +54,21 @@ impl<T: CoordsFloat> CMapBuilder<T> {
     }
 }
 
+macro_rules! if_true_error {
```
cedricchevalier19, line 57: What `if_true_error` means ?

cedricchevalier19 (approved):
Not sure about `if_true_error` is a good name.

PR #95, opened 2024-06-07
doc: update visuals in code doc & user guide

imrn99:
- all changes done in the user guide are in the `Definitions` section
- we need absolute link in the rust doc if we want the figures to appear on docs.rs; its probably possible to prevent that with a custom doc build script, but I won't look into it now
    - affected doc entries: `CMap2`'s example, `CMapBuilder::unit_grid`, `CMapBuilder::unit_triangles`

preview changes:
```shell
# Serve the doc on a local server
mdbook serve --open -d ../target/doc/ user-guide/ &
cargo +nightly doc --all --all-features --no-deps
```

## Scope

- [x] Documentation

## Type of change

- [x] Refactor


honeycomb-core/src/cmapbuilder/grid/mod.rs
```diff
@@ -52,6 +52,17 @@ impl<T: CoordsFloat> CMapBuilder<T> {
 
 impl<T: CoordsFloat> CMapBuilder<T> {
     /// Create a [`CMapBuilder`] with a predefinite [`GridDescriptor`] value.
+    ///
+    /// # Arguments
+    ///
+    /// - `n_square: usize` -- Number of cells along each axis.
+    ///
+    /// # Generated Map
```
dssgabriel, line 60: Wouldn't `Return value` be more explicit?

honeycomb-core/src/cmapbuilder/grid/mod.rs
```diff
@@ -61,6 +72,18 @@ impl<T: CoordsFloat> CMapBuilder<T> {
     }
 
     /// Create a [`CMapBuilder`] with a predefinite [`GridDescriptor`] value.
+    ///
+    /// # Arguments
+    ///
+    /// - `n_square: usize` -- Number of cells along each axis.
+    ///
+    /// # Generated Map
```
dssgabriel, line 80: Same thing here.

dssgabriel (approved):

honeycomb-core/src/cmapbuilder/grid/mod.rs
imrn99: You are probably right, I didn't do it because the returned value is technically a builder struct, but I can add short sentence to clarify that in the section.

PR #96, opened 2024-06-10
feat!(core): add `UnknownAttributeStorage` trait for attribute-agnostic methods

imrn99:
- new trait `UnknownAttributeStorage` which is attribute-agnostic, i.e. isn't defined using `A: AttributeBind`
- move some members of `AttributeStorage` to the new trait
- add index-based `merge` and `split` routines to the new trait in order to process those operations without downcasting the entire storage object

## Scope

- [x] Code

## Type of change

- [x] New feature(s)

## Other

- [x] Breaking change: moved some methods from `AttributeStorage`, add a `From<DartIdentifier>` bound to `AttributeBind::IdentifierType`

## Necessary follow-up

- [x] Needs testing

### DIFF
--- a/honeycomb-core/src/attributes/collections.rs
+++ b/honeycomb-core/src/attributes/collections.rs
@@ -6,7 +6,7 @@
 // ------ IMPORTS
 
 use crate::attributes::traits::AttributeStorage;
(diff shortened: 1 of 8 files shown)


cedricchevalier19 (approved):
Looks clearer.

PR #102, opened 2024-06-11
feat(core): add storage manager to builder

imrn99:
- add support for the new storage manager struct to `CMapBuilder`; This is the last step before adding the manager to the map & wiring it up

## Scope

- [x] Code

## Type of change

- [x] New feature(s)

### DIFF
--- a/honeycomb-core/src/cmapbuilder/io/mod.rs
+++ b/honeycomb-core/src/cmapbuilder/io/mod.rs
@@ -7,7 +7,8 @@
 
 // ------ IMPORTS
 use crate::{
-    BuilderError, CMap2, CMapBuilder, CoordsFloat, DartIdentifier, Vertex2, VertexIdentifier,
+    AttrStorageManager, BuilderError, CMap2, CMapBuilder, CoordsFloat, DartIdentifier, Vertex2,
+    VertexIdentifier,
 };
 use num::Zero;
 use std::collections::BTreeMap;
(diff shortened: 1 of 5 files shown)


honeycomb-core/src/cmapbuilder/io/mod.rs
```diff
@@ -101,7 +102,10 @@ macro_rules! build_vertices {
 ///         - the number of coordinates cannot be divided by `3`, meaning a tuple is incomplete
 ///         - the number of `Cells` and `CellTypes` isn't equal
 ///         - a given cell has an inconsistent number of vertices with its specified cell type
-pub fn build_2d_from_vtk<T: CoordsFloat>(value: Vtk) -> Result<CMap2<T>, BuilderError> {
+pub fn build_2d_from_vtk<T: CoordsFloat>(
+    value: Vtk,
+    mut manager: AttrStorageManager,
+) -> Result<CMap2<T>, BuilderError> {
```
cedricchevalier19, line 108: It looks a little bit weird to have this manager argument. Perhaps it means that an /object/ should provide this facility?

cedricchevalier19 (approved):

honeycomb-core/src/cmapbuilder/io/mod.rs
imrn99: the reason I'm passing it to the builder is that we might want to use the VTK file's attributes sections when building our map. 

I thought about making a config object, but there are limitations to the information I can put in this object (e.g. type information of attributes/storages cannot be stored), so I decided to just pass around the object since we're inside the building routines.

imrn99: added a note in 96ec0aa

PR #106, opened 2024-06-24
test: add coverage for new attribute-related items

imrn99:
- add tests over the items used to implement the generic attribute system
- fix mistakes in implementation highlighted by new tests

## Scope

## Type of change

- [x] Fix
- [x] Testing

### DIFF
--- a/honeycomb-core/src/cmap2/structure.rs
+++ b/honeycomb-core/src/cmap2/structure.rs
@@ -198,7 +198,8 @@ impl<T: CoordsFloat> CMap2<T> {
     ///
     /// - `n_darts: usize` -- Number of darts composing the new combinatorial map.
     /// - `attr_storage_manager: AttrStorageManager` -- Manager structure holding
-    ///   the user-defined attributes.
+    ///   the user-defined attributes. The containers held by the manager should
+    ///   all be empty.
     ///
     /// # Return
     ///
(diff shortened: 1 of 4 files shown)


honeycomb-core/src/cmap2/structure.rs
```diff
@@ -211,8 +211,9 @@ impl<T: CoordsFloat> CMap2<T> {
     #[must_use = "constructed object is not used, consider removing this function call"]
     pub(crate) fn new_with_attributes(
         n_darts: usize,
-        attr_storage_manager: AttrStorageManager,
+        mut attr_storage_manager: AttrStorageManager,
     ) -> Self {
+        attr_storage_manager.extend_storages(n_darts + 1);
```
cedricchevalier19, line 216: Not very clear: what `extend_storages` do and why `n_darts + 1`?

imrn99: I added comments & a little bit of doc on what's expected of the `CMap2::new_with_attributes` constructor in d67a185. Essentially, all storages are of null length before the call to `extend_storages`; The method is used to prepare all storages before they are accessed when manipulating the map. 

This is necessary because all attribute accesses in the `0..n_darts+1` range should be valid according to how we implement embedding (although accessed values may be `None`)

Issue #110, opened 2024-06-27
tracker: grid submersion algorithm


imrn99:
The first algorithm considered for implementation is a 2D, [Morph](https://internationalmeshingroundtable.com/assets/research-notes/imr32/2011.pdf)-inspired, mesh construction algorithm. This issue is used to group and document the implementation process as well as specific considerations / choices.

## Scope

Some considerations:
- evaluating how combinatorial maps perform in the context of this approach
- evaluating the learning curve of combinatorial maps usage
- ensuring basic correctness & viability before moving on to 3D or parallelization

Some non-considerations (ATM, possibly ever):
- correct capture of highly contrasted (size-wise) geometry features requiring some form of adaptative mesh refinement
- significant work on algorithm writing; our focus is on data structures

### Hypotheses

-  All components of the geometry are located in positive X/Y
- Edges are consistently oriented (i.e. normals of edges making up a face all point outward / inward, no mix)

### Map generation

**Input**:
- Geometry of interest
- Grid characteristics (length of cell along X/Y-axis)
 
**Algorithm**:

Let `S` be the set of all segments making up the boundaries of the geometry.
Let `PoI` be the set of points of interests of the boundaries (i.e. vertices that must be inserted into the final mesh)
Let `DoI` be the corresponding set of darts of interests
 
For all segments `[A, B]` of `S`: 
- compute the Manhattan distance `d` between cell(A) and cell(B):
    - if  `d == 0`: `A` and `B` belong to the same grid cell
        - Do nothing
    - if `d == 1`: `A` and `B` belong to neighbor cells
        - Compute the intersection `C` between the segment and the grid's edge
        - Split the grid's edge to add `C` on it, add relevant darts to `DoI`
        - Replace segment `[A, B]` by segments `[A, C]`, `[C, B]`
        - Add `C` to `PoI`
    - if `d > 1`: `A` and `B` belong to different, non-neighbor cells
        - Compute all intersections `Ci` between the segment and grid's edges
        - Split grid's edges to add `Ci`s on them, add relevant darts to `DoI`
        - Replace segment `[A, B]` by segments `[A, C1]`, `[C1, C2]`, ..., `[CX, B]`
        - Add all `Ci`s to `PoI`
- if `B` belongs to `PoI`:
    - Insert `B` into the map
    - Add relevant darts to `DoI`

For all points `P` of `PoI`:
- search `S` to find `P'`, the first "next" point to belong to `PoI`
[...]

PR #113, opened 2024-07-03
feat(kernels): implement pre-processing components of `grisubal`

imrn99:
part of #110 

implement a few routines (bounding box computation, Manhattan distance) & adjust some signatures.

## Scope

- [x] Code

## Type of change

- [x] New feature(s)

### DIFF
--- a/honeycomb-kernels/src/grisubal/inp.rs
+++ b/honeycomb-kernels/src/grisubal/inp.rs
@@ -14,7 +14,7 @@ use vtkio::Vtk;
 
 /// Post-processing clamp operation.
 #[derive(Default)]
-pub enum Clamp {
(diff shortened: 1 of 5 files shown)


honeycomb-kernels/src/grisubal/inp.rs
```diff
@@ -14,7 +14,7 @@ use vtkio::Vtk;
 
 /// Post-processing clamp operation.
 #[derive(Default)]
-pub enum Clamp {
+pub enum Clip {
```
cedricchevalier19, line 17: Update comments?

imrn99: Oops, I missed those

PR #114, opened 2024-07-11
feat: implement geometry reading from vtk input

imrn99:
part of #110 

## Scope

- [x] Code

## Type of change

- [x] New feature(s)

### DIFF
--- a/honeycomb-kernels/Cargo.toml
+++ b/honeycomb-kernels/Cargo.toml
@@ -15,3 +15,4 @@ publish = true
 [dependencies]
 honeycomb-core = { workspace = true, features = ["io", "utils"] }
 vtkio.workspace = true
+num.workspace = true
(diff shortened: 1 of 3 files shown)


honeycomb-kernels/Cargo.toml
```diff
@@ -15,3 +15,4 @@ publish = true
 [dependencies]
 honeycomb-core = { workspace = true, features = ["io", "utils"] }
 vtkio.workspace = true
+num.workspace = true
```
cedricchevalier19, line 18: What is this workspace?

imrn99: The actual version of dependencies are defined in the root manifest (i.e. the workspace's); This is basically telling cargo to use the version specfied in the workspace, that way there is only one `num` dep to update even if multiple member use the crate.

PR #116, opened 2024-07-17
feat(kernel): implement the main `grisubal` routine

imrn99:
This PR contains the implementation of the main routine of the `grisubal` algorithm, i.e. the capture of the input geometry into a `CMap` via *grid interpolation*.

Part of #110 

## Scope

- [x] Code: `honeycomb-kernels`

## Type of change

- [x] New feature(s)

## Necessary follow-up

- [x] Needs documentation
- [x] Needs testing
- [x] Other: the main routine is almost 1K lines long, it might be a good idea to extract some functions / macros for readability & maintainability


honeycomb-kernels/src/grisubal/kernel.rs
```diff
@@ -24,8 +58,746 @@ use honeycomb_core::{CMap2, CoordsFloat};
...
+    clippy::cast_sign_loss
+)]
 pub fn build_mesh<T: CoordsFloat>(geometry: &Geometry2<T>, grid_cell_sizes: (T, T)) -> CMap2<T> {
-    todo!()
+    // build the overlapping grid we'll modify
+    let bbox = geometry.bbox();
+    let (cx, cy) = grid_cell_sizes; // will need later
+    let (nx, ny) = (
+        (bbox.max_x / cx).ceil().to_usize().unwrap(),
+        (bbox.max_y / cy).ceil().to_usize().unwrap(),
```
cedricchevalier19, line 73: `min_x` and `min_y` are 0?

honeycomb-kernels/src/grisubal/kernel.rs
```diff
@@ -24,8 +58,746 @@ use honeycomb_core::{CMap2, CoordsFloat};
...
+                    // cross up
+                    ( 0,  1) => {
+                        let dart_id = d_base + 2;
+                        // vertex associated to crossed dart, i.e. up right corner
+                        let v_dart = cmap.vertex(cmap.vertex_id(dart_id)).unwrap();
+                        let (_, t) = up_intersec!(v1, v2, v_dart, cx);
+                        (t, dart_id)
+                    }
+                    _ => unreachable!(),
+                };
```
cedricchevalier19, line 171: Try to factorize the code.

cedricchevalier19 (changes requested):
Please, introduce auxiliary functions to decrease the code complexity. The algorithm is hard to distinguish. Most of the lines are repeated implementation details that we do not really want to see looking at the algorithm.

I'll approve once the algorithm will be split into functions.

However, I think it should be useful to shorten the code in another PR: I do not fully understand the logic of treating all the cases that are constrained cases of the general case treated at the end, without really going back to the simpler cases.

honeycomb-kernels/src/grisubal/kernel.rs
imrn99: They currently are hardcoded to 0, yes. This should not create errors since we require the geometry to be defined in positive Xs/Ys domain, but this could be easily improved with the rewrite I've done in the latest commits.

imrn99:
> Please, introduce auxiliary functions to decrease the code complexity. The algorithm is hard to distinguish. Most of the lines are repeated implementation details that we do not really want to see looking at the algorithm.
>
> I'll approve once the algorithm will be split into functions.

I've separated the steps into functions & rewrote some part. This also led to ~300 less lines of code, so it should be much easier to read, even if the split is very rough.

> However, I think it should be useful to shorten the code in another PR: I do not fully understand the logic of treating all the cases that are constrained cases of the general case treated at the end, without really going back to the simpler cases.

Extracting more function / making the execution flow more uniform would require implementation choices that probably deserve a discussion. We can probably deal with that in person.

cedricchevalier19 (approved):
Thanks for the last updates, code is a lot more readable.

PR #119, opened 2024-07-23
refactor(kernels): adjust clipping policy

imrn99:
Summary of the changes if needed.

## Scope

- [x] Code

## Type of change

- [x] Refactor


cedricchevalier19 (commented):
I am not sure about `normal` and `antinormal`.

I do not see the link with orthogonality, and to describe the two sides, `left` and `right` (or `positive/negative`) are more explicit in the long run.
Positive/negative or `direct/indirect` might be better to extend it to 3D.

@franck-ledoux Any thoughts?

imrn99:
I renamed the variants to `Left` and `Right`; The CI error is expected (& fixed in another PR)

PR #121, opened 2024-07-24
fix(render): add condition over shrink dir to prevent some crashes

imrn99:
this cover crashes in case at least one vertex of a face is on its center of gravity


dssgabriel (approved):
Newb question but will this just panic at the `unwrap` if not returning the `Ok(_)` variant?

cedricchevalier19:
> Newb question but will this just panic at the `unwrap` if not returning the `Ok(_)` variant?

Yes, it used to panic when not `Ok`, now it is just skipped.

imrn99:
> Newb question but will this just panic at the `unwrap` if not returning the `Ok(_)` variant?

It won't, this would be somewhat equivalent to:

```rust
// res: Result<T, E>
if res.is_ok() {
    let val = res.unwrap(); // val: T
    // ...
}
```

PR #124, opened 2024-07-25
refactor!(core): rewrite `Vertex2` & `Vector2` as tuple structs

imrn99:
- rewrite both `Vertex2` & `Vector2` as tuple struct, i.e. `struct Vector2(T, T);`
- delete `Coords2` struct (unused after rewrite)

This results in significantly less boilerplate when using these structs, most notably when building instances

```rust
// before
let v1 = Vertex2::from((1.0, 2.0)); // init
assert_eq!(v1.x(), 1.0);            // named access
assert_eq!(v1[1], 2.0);             // index-based access - inner impl uses an assertion to check index

// after
let v1 = Vertex2(1.0, 2.0);   // init - much better
assert_eq!(v1.x(), 1.0);      // named access - unchanged
assert_eq!(v1.1, 2.0);        // index-based access - better, and prevent indexing with incoherent values
```

## Scope

- [x] Code: changes in implementation affect the core crate; other changes are just adjustment to usages

## Type of change

- [x] Refactor

## Other

- [x] Breaking change


cedricchevalier19 (approved):
Indeed, code is shorter and more idiomatic.
Good job!

PR #125, opened 2024-07-25
chore: move asset files to a dedicated submodule

imrn99:
This should help us to keep a clean repo in the long term

## Scope

- [x] Repository

## Type of change

- [x] Chore


imrn99:
Closing this as @cedricchevalier19 suggested that there might be smarter way to do this. We can also look into github LFS when we'll deal with larger meshes (i.e. benchmarking instead of debugging).

PR #128, opened 2024-07-26
feat(core): add `remove_storage` method to `AttrStorageManager`

imrn99:
Add a method that allows removing a given storage from the manager entirely. This will be useful for the `grisubal` implementation (see #110), more specifically to remove the attribute used to model clipping regions.

## Scope

- [x] Code: `honeycomb-core`

## Type of change

- [x] New feature(s)

### DIFF
--- a/honeycomb-core/src/attributes/manager.rs
+++ b/honeycomb-core/src/attributes/manager.rs
@@ -428,6 +428,24 @@ impl AttrStorageManager {
             .expect("E: could not downcast generic storage to specified attribute type")
     }
 
+    /// Remove an entire attribute storage from the manager.
+    ///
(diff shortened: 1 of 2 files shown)


cedricchevalier19 (commented):
What happens to the attributes? Is it possible to still reference them after removal?

imrn99:
> What happens to the attributes? Is it possible to still reference them after removal?

After removal, the storage will behave as if the attribute never existed. For example, if the user tries to access a value after, the program will crash with the error message `E: could not find storage associated to the specified attribute type`.

If you can think of a use case where we would want to get the removed storage, I can modify the method to return it.

cedricchevalier19 (approved):

PR #129, opened 2024-07-29
feat(kernels): add a dedicated intersection insertion step

imrn99:
This PR modify step 1 and step 2 of the algorithm to insert a new step in between. This step ensures correct vertex insertion even if a single edge is intersected multiple times on both of its sides.

As a side note, the `grisubal` example is now executed on the input file passed as argument: 

```
cargo run  --example grisubal -- path/to/file.vtk
```

## Scope

- [x] Code

## Type of change

- [x] Refactor

## Necessary follow-up

- [x] Needs documentation
- [x] Other: This step can be used to process specifically tangents & corner intersections

### DIFF
--- a/examples/examples/kernels/grisubal.rs
+++ b/examples/examples/kernels/grisubal.rs
@@ -1,16 +1,24 @@
 use honeycomb_kernels::grisubal;
 use honeycomb_render::{RenderParameters, SmaaMode};
 
+use std::env;
+
 fn main() {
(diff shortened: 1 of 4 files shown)


imrn99:
CI fails because `rand` was updated one hour ago, and the version we use isn't recognized properly by cargo @cedricchevalier19

cedricchevalier19 (approved):
Perhaps you could replace `WARNING` with `FIXME`

PR #135, opened 2024-07-30
refactor(kernels): ignore points of interest landing exactly on the grid

imrn99:
If a PoI is located exactly on the grid, it should be captured by intersection. To prevent duplicating vertices, we can remove the vertex's PoI status & capture the vertex using the regular intersection case.

## Scope

- [x] Code: `honeycomb-kernels`

## Type of change

- [x] Refactor

### DIFF
--- a/honeycomb-kernels/src/grisubal/kernel.rs
+++ b/honeycomb-kernels/src/grisubal/kernel.rs
@@ -125,6 +125,17 @@ pub fn build_mesh<T: CoordsFloat>(geometry: &Geometry2<T>, grid_cell_sizes: (T, 
 
 // --- main kernels steps
 
+fn remove_redundant_poi<T: CoordsFloat>(geometry: &mut Geometry2<T>, (cx, cy): (T, T)) {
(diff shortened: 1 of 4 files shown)


honeycomb-kernels/src/grisubal/kernel.rs
```diff
@@ -125,6 +132,18 @@ pub fn build_mesh<T: CoordsFloat>(geometry: &Geometry2<T>, grid_cell_sizes: (T,
 
 // --- main kernels steps
 
+fn remove_redundant_poi<T: CoordsFloat>(geometry: &mut Geometry2<T>, (cx, cy): (T, T)) {
```
cedricchevalier19, line 135: I do not see this function as a `main kernel step`. Perhaps it should be moved in another part of the file.

cedricchevalier19 (approved):
Two comments:

1. Can you rename the PR with "points of interest" instead of "PoI"?
2. Add a comment that this function only work on a regular cartesian grid.

PR #136, opened 2024-07-31
feat(kernels): add new intersection type dedicated to corners

imrn99:
Add `GeometryVertex::IntersecCorner` variant to handle segments that cross grid cells across corners (not edges). In this case, the vertex insertion logic is bypassed and we directly save a `DartIdentifier` that can be used to build `MapEdge`s at step 3.

## Scope

## Type of change

- [x] New feature(s)


cedricchevalier19 (commented):
Intersect with a T looks like a better name, what do you think about renaming your function?

imrn99:
> Intersect with a T looks like a better name, what do you think about renaming your function?

Is `generate_intersection_data` acceptable?

PR #137, opened 2024-08-01
feat(kernels::grisubal): mark boundary darts during map construction

imrn99:
use attributes to mark boundaries of the captured geometry during construction

the attribute `Boundary` binds to dart and can take 3 different values according to the dart's location: left side of the boundary, right side of the boundary, or not on the boundary 

## Scope

## Type of change

- [x] New feature(s)


honeycomb-kernels/src/grisubal/model.rs
```diff
@@ -237,28 +237,37 @@ pub struct MapEdge<T: CoordsFloat> {
...
 }
 
-#[derive(Clone, Copy, Debug)]
-pub struct IsBoundary(bool);
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum Boundary {
+    Left,
+    Right,
+    None,
+}
```
cedricchevalier19, line 245: Can you comment on what it means?

imrn99: I've added comments & a figure in 29aac32

cedricchevalier19 (approved):
LGTM

PR #140, opened 2024-08-02
feat(kernels::grisubal): detect some boundary orientation issues

imrn99:
This PR adds a function that detect orientation consistency **per boundary** of a given `Geometry2`. **per boundary** means that inconsistencies between two different boundaries (e.g. a geometry with a hole in it) won't be detected by this function.

This PR also adds an error enum specific to `grisubal`, `GrisubalError`.

## Scope

## Type of change

- [x] New feature(s)


honeycomb-kernels/src/grisubal/model.rs
```diff
@@ -201,6 +203,37 @@ impl<T: CoordsFloat> From<Vtk> for Geometry2<T> {
...
+    geometry: &Geometry2<T>,
+) -> Result<(), GrisubalError> {
+    let mut origins = HashSet::new();
+    let mut endpoints = HashSet::new();
+
+    for (orig, endp) in &geometry.segments {
+        if !origins.insert(orig) {
+            return Err(GrisubalError::InconsistentOrientation(
+                "two segments have the same vertex as their respective origin",
+            ));
```
cedricchevalier19, line 225: Can you add in the message the identifier that is detected as duplicated?

imrn99: I've done that in 47da67f, but it requires changing the type of the enum variant from `&'static str` to `String`

PR #156, opened 2024-08-30
refactor!(examples): filter provided examples

imrn99:
There were too many examples, most of which were redundant. I removed a bunch and added back simplified versions.

## Scope

## Type of change

- [x] Refactor

## Other

- [x] Breaking change


cedricchevalier19 (approved):
Looks OK.

PR #163, opened 2024-09-06
refactor!(core): replace `from_ *`function of `CMapBuilder` by trait impls

imrn99:
replace `from_*` functions of `CMapBuilder` by `impl From<*>`

## Scope

- [x] Code

## Type of change

- [x] Refactor

## Other

- [x] Breaking change

### DIFF
--- a/honeycomb-core/src/cmap/builder/grid/mod.rs
+++ b/honeycomb-core/src/cmap/builder/grid/mod.rs
@@ -25,25 +25,25 @@ impl<T: CoordsFloat> CMapBuilder<T> {
         self.grid_descriptor = Some(grid_descriptor);
         self
     }
-
-    #[cfg(feature = "utils")]
-    /// Create a [`CMapBuilder`] from a [`GridDescriptor`].
-    ///
-    /// This function is roughly equivalent to the following:
-    ///
-    /// ```rust
-    /// # use honeycomb_core::prelude::{CMapBuilder, GridDescriptor};
-    /// // setup grid descriptor
-    /// let gridd = GridDescriptor::default();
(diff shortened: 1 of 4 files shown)


honeycomb-core/src/cmap/builder/grid/mod.rs
```diff
@@ -72,7 +72,7 @@ impl<T: CoordsFloat> CMapBuilder<T> {
         let gridd = GridDescriptor::default()
             .n_cells([n_square; 3])
             .len_per_cell([T::one(); 3]);
-        CMapBuilder::from_grid_descriptor(gridd)
+        CMapBuilder::from(gridd)
```
cedricchevalier19, line 75: Can we use `into()` here? We already know that we will return a `CMapBuilder`.

honeycomb-core/src/cmap/builder/grid/mod.rs
```diff
@@ -97,7 +97,7 @@ impl<T: CoordsFloat> CMapBuilder<T> {
             .n_cells([n_square; 3])
             .len_per_cell([T::one(); 3])
             .split_quads(true);
-        CMapBuilder::from_grid_descriptor(gridd)
+        CMapBuilder::from(gridd)
```
cedricchevalier19, line 100: Same here.

cedricchevalier19 (approved):

honeycomb-core/src/cmap/builder/grid/mod.rs
imrn99: Yes we can, nice catch

PR #164, opened 2024-09-06
feat(bench, kernels): setup grisubal bench tools

imrn99:
This setup includes:

- a binary calling the grisubal function once; this is useful for perf / heaptrack / flamegrahes
- a `profiling` feature for the kernel crate; for now, this enables time measurements per section
- criterion benchmarks for execution time, and grid size variation; note that input cannot be passed via CLI for those

This is a very crude setup, but it can be used to guide first optimizations and profile the kernel

### Usage Example

You can first build the binary:

```
cargo build --bin=grisubal --release --features profiling
```

Then run the program, it takes a path to a vtk file as argument. For example, using this [repo](https://github.com/imrn99/meshing-samples):

```
./target/release/grisubal ~/Repositories/meshing-samples/vtk/2D/C5_2D.vtk
```

This will print out something similar to this; times are in nanoseconds:

```
164417, 58042, 15667, 12834, 1542, 1168167, 825666, 152083, 150375, 8166, 29000, 42, 1459
```

It's possible to generate a `csv` file by writing a small script that outputs this line into a file. From there, we can easily compute averages / relative parts / ... . Note that we expect the input not to generate any warnings since these would disrupt the `csv` file.

## Scope

- [x] Code: both the benches and kernels crates

## Type of change

- [x] New feature(s)


benches/benches/grisubal/grid_size.rs
```diff
@@ -0,0 +1,47 @@
...
+};
+use honeycomb_benches::FloatType;
+
+// ------ CONTENT
+
+pub fn criterion_benchmark(c: &mut Criterion) {
+    // passing args to cargo bench filters bench instead of actually reading args;
+    // the path to the file needs to be hardcoded (I think?)
+    // let args: Vec<String> = std::env::args().collect();
+    let path = "path/to/file.vtk";
```
cedricchevalier19, line 16: How does it work?

cedricchevalier19 (commented):
Can you add an example output to the PR description?

benches/benches/grisubal/grid_size.rs
imrn99: I added a comment to the line. Basically, this str should be replaced with a path to an actual input by whoever runs the benchmark. 

An alternative would be to add whatever input we want to benchmark in the repo, so that we can add the correct path.

imrn99:
I added an example of a workflow to the PR's description. I could also add it to the doc or the user guide if necessary

benches/benches/grisubal/grid_size.rs
cedricchevalier19: Put at least a file in the repo to make sure everything is working without having to modify the files.

imrn99: it's done!

PR #167, opened 2024-09-12
refactor!(core): remove `Orbit2::is_isolated`

imrn99:
remove this method since it's unused; I don't think keeping it for users is a good idea since it is pretty costly

## Scope

- [x] Code

## Type of change

- [x] Refactor

## Other

- [x] Breaking change


cedricchevalier19 (approved):
I agree, the removal makes sense.

If a need comes later, you will still able to reach the code from git history.

PR #169, opened 2024-09-13
refactor(core): improve some error enums

imrn99:
- add a variant to `CoordsError` to distinguish failure in unit vector computation vs normal vector computation
- remove `ManagerError`; panicking is probably fine since the error is thrown at init, not during the execution of an algorithm

## Scope

- [x] Code

## Type of change

- [x] Refactor

## Other

- [x] Breaking change

## Necessary follow-up

- [x] Other: the `CMapError` should also be removed, but it will result in significant changes, so it will be in another PR

### DIFF
--- a/honeycomb-core/src/attributes/manager.rs
+++ b/honeycomb-core/src/attributes/manager.rs
@@ -11,13 +11,6 @@ use std::{any::TypeId, collections::HashMap};
 
 // ------ CONTENT
 
-/// Attribute manager error enum.
-#[derive(Debug)]
-pub enum ManagerError {
-    /// Storage of a given type already exists in the structure.
-    DuplicateStorage,
-}
-
 /// Main attribute storage structure.
 ///
 /// This structure is used to store all generic attributes that the user may add to the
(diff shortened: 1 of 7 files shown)


honeycomb-core/src/attributes/manager.rs
```diff
@@ -381,9 +368,10 @@ impl AttrStorageManager {
         }
         .is_some()
         {
-            Err(ManagerError::DuplicateStorage)
-        } else {
-            Ok(())
+            panic!(
+                "E: Storage of attribute `{}` already exists in the attribute storage manager",
+                std::any::type_name::<A>()
+            );
```
cedricchevalier19, line 374: I am not a massive fan of `panic!` in a library.

You can leave it if you think the user cannot recover this error, but I think it is a regression there.

cedricchevalier19 (commented):
I understand the aim to have a more straightforward API but I think panicking should be avoided in a library.

honeycomb-core/src/attributes/manager.rs
imrn99: I switched it to a warning in d86f664. That's a good point about panic in libraries; I'll try to go over all of those in a separate PR

PR #171, opened 2024-09-13
refactor!(core): remove `CMapError`

imrn99:
This enum had a single variant: `CMapError::UndefinedVertex`. Removing the error and replacing `Result`s with `Option`s simplifies the code and makes the vertex/attribute metho signatures consistent. 

important changes are in the `honeycomb-core/src/cmap/dim2/embed.rs` file

## Scope

- [x] Code: core crate, other changes are just fixes 

## Type of change

- [x] Refactor

## Other

- [x] Breaking change


cedricchevalier19 (approved):
This is a significant change in semantics, but I see why looking for a non-existing member in a map can be expected.

PR #172, opened 2024-09-16
feat(core): implement attribute storage traits for `std` collections

imrn99:
- rename some methods of the traits to avoid name conflicts
- implement `AttributeStorage` & `UnknownAttributeStorage` for `Vec<Option<A>>`

## Scope

- [x] Code

## Type of change

- [x] New features

## Other

- [x] Breaking change

### DIFF
--- a/honeycomb-core/src/attributes/collections.rs
+++ b/honeycomb-core/src/attributes/collections.rs
@@ -11,6 +11,67 @@ use num_traits::ToPrimitive;
 
 // ------ CONTENT
 
+impl<A: AttributeBind + AttributeUpdate + Copy> UnknownAttributeStorage for Vec<Option<A>> {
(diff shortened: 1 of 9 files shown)


imrn99:
@cedricchevalier19 I'm not sure if those changes are good. The idea was to simplify storage types, allowing users to store custom attributes in standard `Vec<Option>`. If this is kept, I would delete the struct `AttrSparseVec`

cedricchevalier19:
I think you are right, this one is tricky. We need to discuss it quite a bit. We can do it tomorrow.

imrn99:
Closing this; implementing our traits on `std` collections just creates confusion; having abstractions is also useful if we want to change the underlying storage without breaking everything

PR #173, opened 2024-09-16
refactor!(core): remove lifetime from `OrbitPolicy` & update `AttributeBind`

imrn99:
- remove arbitrary lifetime from `OrbitPolicy`
- replace `AttributeBind::binds_to()` with a const `OrbitPolicy`

I don't think we need runtime-valued orbit policy, & this simplifies the struct

## Scope

- [x] Code

## Type of change

- [x] Refactor

## Other

- [x] Breaking change


cedricchevalier19 (approved):
Ok to consider that `OrbitPolicy` does not depend on a `CMap2` instance.

Issue #175, opened 2024-09-17
enh: Error handling


cedricchevalier19:
Discuss and design a coherent error handling.

## Information

- [x] Enhancement of existing content
- [ ] Suggestion for a new feature
- [ ] Other: ...

### Current state

There are:
- some custom `Error` types
- some "warnings" 
- some `panic` or `unimplemented`

See #169, #174 

### Proposal

I'd appreciate it if you could find a coherent way to deal with different errors. It should be lightweight, but it should be thought out thoroughly to make honeycomb usable as a library.

imrn99:
good point for usability. I'll use this issue to gather relevant data

imrn99:
# explicit panics, expects

Ignoring unreachable panics.

## panics

```
honeycomb-core/src/cmap/builder/io/tests.rs:32:  _ => panic!("cmap was built incorrectly"),
honeycomb-core/src/cmap/builder/io/mod.rs:27:    Vtk::import(file_path).unwrap_or_else(|e| panic!("E: failed to load file: {e:?}"));
honeycomb-core/src/cmap/builder/io/mod.rs:49:    Vtk::import(value).unwrap_or_else(|e| panic!("E: failed to load file: {e:?}"));
honeycomb-kernels/src/grisubal/mod.rs:144:       Err(e) => panic!("E: could not open specified vtk file - {e}"),
```

## expects

```
honeycomb-core/src/attributes/collections.rs:65:         .expect("E: cannot split attribute value - value not found in storage");
honeycomb-core/src/attributes/collections.rs:179:        .expect("E: cannot split attribute value - value not found in storage");
honeycomb-core/src/attributes/manager.rs:65:///          .expect("E: could not downcast generic storage to specified attribute type")
honeycomb-core/src/attributes/manager.rs:320:            .expect("E: could not find storage associated to the specified attribute type")
honeycomb-core/src/attributes/manager.rs:322:            .expect("E: could not downcast generic storage to specified attribute type");
honeycomb-core/src/attributes/manager.rs:335:            .expect("E: could not find storage associated to the specified attribute type")
honeycomb-core/src/attributes/manager.rs:337:            .expect("E: could not downcast generic storage to specified attribute type");
honeycomb-core/src/attributes/manager.rs:414:            .expect("E: could not downcast generic storage to specified attribute type")
honeycomb-core/src/cmap/builder/io/mod.rs:125:           .expect("failed to load piece data - is it not inlined?");
[...]

PR #176, opened 2024-09-19
refactor: replace `unwrap` calls with more explicit alternatives

imrn99:
To build a more consistent error handling logic (see #175), I repalced a number of `unwrap` calls with either:
- `expect` calls, to clarify what went wrong
- a `Result` return, if the error was not something that should completely crash the program

a lot of the unwraps were unreachable due to previous checks in execution. the memory methods won't crash on write failure; they return a result which can be processed by the user.

## Scope

- [x] Code

## Type of change

- [x] Refactor


cedricchevalier19 (approved):
It is some progress but on the longer run we have to avoid any premature return from the library.

PR #177, opened 2024-09-20
feat(kernels): implement polygon triangulation routines

imrn99:
Implement two polygon triangulation methods:
- ear clipping, which works on all polygons with no hole
- fanning, which only works on a subset of polygons; there is a generic implementation of this method and one dedicated to convex polygons

## Scope

- [x] Code

## Type of change

- [x] New feature(s)

### DIFF
--- a/honeycomb-kernels/src/triangulation/ear_clipping.rs
+++ b/honeycomb-kernels/src/triangulation/ear_clipping.rs
@@ -1,4 +1,4 @@
-use honeycomb_core::cmap::{CMap2, DartIdentifier, FaceIdentifier};
+use honeycomb_core::cmap::{CMap2, DartIdentifier, FaceIdentifier, Orbit2, OrbitPolicy};
 use honeycomb_core::geometry::CoordsFloat;
 
 pub fn process_cell<T: CoordsFloat>(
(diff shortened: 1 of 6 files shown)


honeycomb-kernels/src/triangulation/ear_clipping.rs
```diff
@@ -0,0 +1,134 @@
...
+) {
+    let mut n = Orbit2::new(cmap, OrbitPolicy::Custom(&[1]), face_id as DartIdentifier).count();
+
+    // early rets
+    if n <= 3 {
+        println!("I: face {face_id} is an {n}-gon -- skipping triangulation");
+        return;
+    }
+    if (n - 3) * 2 != new_darts.len() {
+        println!("W: not enough pre-allocated darts to triangulate face {face_id} -- skipping triangulation");
```
cedricchevalier19, line 57: Is it always "not enough"?

honeycomb-kernels/src/triangulation/fan.rs
```diff
@@ -0,0 +1,176 @@
...
+    let n = darts.len();
+
+    // early rets
+    if n <= 3 {
+        println!("I: face {face_id} is an {n}-gon -- skipping triangulation");
+        return;
+    }
+    if (n - 3) * 2 != new_darts.len() {
+        println!("W: not enough pre-allocated darts to triangulate face {face_id} -- skipping triangulation");
+        return;
```
cedricchevalier19, line 49: Same here.

honeycomb-kernels/src/triangulation/fan.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    let vertices: Vec<_> = darts
+        .iter()
+        .map(|dart_id| {
+            cmap.vertex(cmap.vertex_id(*dart_id))
+                .expect("E: found a topological vertex with no associated coordinates")
+        })
+        .collect();
```
cedricchevalier19, line 58: Duplicated code with `ear_clipping.rs`. Does a function make sense?

honeycomb-kernels/src/triangulation/fan.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    // early rets
+    if n <= 3 {
+        println!("I: face {face_id} is an {n}-gon -- skipping triangulation");
+        return;
+    }
+    if (n - 3) * 2 != new_darts.len() {
+        println!("W: not enough pre-allocated darts to triangulate face {face_id} -- skipping triangulation");
+        return;
+    }
```
cedricchevalier19, line 153: Same here.

cedricchevalier19 (commented):
Nothing is blocking the PR, but I think you can improve the code before merging.

imrn99:
@cedricchevalier19 I added three functions:
- `check_requirements`: this takes care of some of the duplicated code (number of vertex per face, number of pre-allocated darts)
- `fetch_face_vertices`: if all vertices of a face are defined, return a collection with those, else raise en error
- `crossp_from_verts`: this replace the macro; it should be inlined anyway

I also added the `TriangulateError` to describe these issues. The routines now return a `Result<(), TriangulateError>` that can be processed to shut down the program or attempt triangulation with another method.

cedricchevalier19 (approved):
LGTM

PR #180, opened 2024-09-27
refactor!(core): change `split_edge` & `splitn_edge` to use preallocated darts

imrn99:
both methods now take a slice of dart IDs to use as new segments components. both methods check if:
- the number of darts is consistent
- each dart is free
- each dart is non-null

## Scope

- [x] Code

## Type of change

- [x] Refactor

## Other

- [x] Breaking change

### DIFF
--- a/honeycomb-core/src/cmap/dim2/advanced_ops.rs
+++ b/honeycomb-core/src/cmap/dim2/advanced_ops.rs
@@ -50,7 +50,7 @@ impl<T: CoordsFloat> CMap2<T> {
     ///  1             2    =>    1      3      2   | + denote darts that encode vertex IDs
     ///    <----2----+              <-4-- <-2-+     |
     /// ```
-    pub fn split_edge(
+    pub fn split_edge_noalloc(
         &mut self,
         edge_id: EdgeIdentifier,
         new_darts: (DartIdentifier, DartIdentifier), // 2D => statically known number of darts
@@ -176,7 +176,7 @@ impl<T: CoordsFloat> CMap2<T> {
     /// // split
     /// let nds = map.add_free_darts(6);
     /// let new_darts: Vec<_> = (nds..nds+6).collect();
-    /// map.splitn_edge(1, &new_darts ,&[0.25, 0.50, 0.75]);
+    /// map.splitn_edge_no_alloc(1, &new_darts ,&[0.25, 0.50, 0.75]);
     /// // after
     /// //    <-<-<-<
     /// //  1 -3-4-5- 2
@@ -201,7 +201,7 @@ impl<T: CoordsFloat> CMap2<T> {
     /// assert_eq!(map.beta::<2>(4), 6);
     /// assert_eq!(map.beta::<2>(5), 2);
     /// ```
-    pub fn splitn_edge(
+    pub fn splitn_edge_no_alloc(
         &mut self,
         edge_id: EdgeIdentifier,
         new_darts: &[DartIdentifier],
--- a/honeycomb-core/src/cmap/dim2/advanced_ops.rs
+++ b/honeycomb-core/src/cmap/dim2/advanced_ops.rs
@@ -261,38 +261,28 @@ impl<T: CoordsFloat> CMap2<T> {
     /// This implementation is 2D specific.
     /// </div>
     ///
-    /// This method takes all darts of an edge and rebuild connections in order to create a new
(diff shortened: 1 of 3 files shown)


honeycomb-core/src/cmap/dim2/advanced_ops.rs
```diff
@@ -50,18 +50,31 @@ impl<T: CoordsFloat> CMap2<T> {
     ///  1             2    =>    1      3      2   | + denote darts that encode vertex IDs
     ///    <----2----+              <-4-- <-2-+     |
     /// ```
-    pub fn split_edge(&mut self, edge_id: EdgeIdentifier, midpoint_vertex: Option<T>) {
+    pub fn split_edge(
+        &mut self,
+        edge_id: EdgeIdentifier,
+        new_darts: (DartIdentifier, DartIdentifier), // 2D => statically known number of darts
+        midpoint_vertex: Option<T>,
+    ) {
```
cedricchevalier19, line 58: Update the documentation above with the new prototype.

It is not clear what `new_darts` is, and why the user has to pass these ids.

cedricchevalier19 (changes requested):
Can you comment more on what is the need here?

The new API looks harder to use than the existing one.

If it is for performance, I think you should leave the old interface and add a new method like `split_edge_noalloc` to allow optimization in some algorithms.

honeycomb-core/src/cmap/dim2/advanced_ops.rs
imrn99: I've added details about dart IDs in the documentation (c9b7ec9)

honeycomb-core/src/cmap/dim2/advanced_ops.rs
```diff
@@ -252,3 +252,263 @@ impl<T: CoordsFloat> CMap2<T> {
...
+    /// # Arguments
+    ///
+    /// - `edge_id: EdgeIdentifier` -- Edge to split in two.
+    /// - `new_darts: (DartIdentifier, DartIdentifier)` -- Dart IDs used to build the new segments.
+    /// - `midpoint_vertex: Option<T>` -- Relative position of the new vertex, starting from the
+    ///   vertex of the dart sharing `edge_id` as its identifier.
+    ///
+    /// ## Dart IDs Requirements & Usage
+    ///
+    /// Because of the dimension, the number of dart needed to perform this operation is at most
```
imrn99, line 282: Explanations are here; I can add more details if things are unclear

honeycomb-core/src/cmap/dim2/advanced_ops.rs
```diff
@@ -252,3 +252,263 @@ impl<T: CoordsFloat> CMap2<T> {
...
+    /// # Arguments
+    ///
+    /// - `edge_id: EdgeIdentifier` -- Edge to split in two.
+    /// - `new_darts: &[DartIdentifier]` -- Dart IDs used to build the new segments.
+    /// - `midpoint_vertices: &[T]` -- Relative positions of new vertices, starting from the
+    ///   vertex of the dart sharing `edge_id` as its identifier.
+    ///
+    /// ## Dart IDs Requirements & Usage
+    ///
+    /// Because of the dimension, we can easily compute the number of dart needed to perform this
```
imrn99, line 406: same here

cedricchevalier19 (commented):
Are the `no_alloc` and the old version not somehow factorizable?

Otherwise, do you have a benchmark to show interest in the new approach?

imrn99:
@cedricchevalier19 

1. I can definitely define an internal function with common code to all of these methods
2. I don't have benchmark, but I have a setup in #164. What I can do is reverse changes in the kernel implementation (i.e. keep original methods) & merge both PRs. After this, I can benchmark the algorithm with both methods, looking at exec time and number of allocator calls

cedricchevalier19:
For 2, only if it is not too much work.
Is performance the only expected gain, or do you think it can have other positive outcome, like better ordering for the ids?

imrn99:
> For 2, only if it is not too much work. 

I don't think it would take too much time, at least checking the number of allocator calls wouldn't

> Is performance the only expected gain, or do you think it can have other positive outcome, like better ordering for the ids?

In sequential, there wouldn't be any change in ID ordering, but in parallel it would guarantee that all new darts are packed together. Original methods added darts as needed while new methods require a slice.

PR #181, opened 2024-09-27
feat(core): add 3D geometry primitives

imrn99:
- add `Vector3` and related ops
- add `Vertex3` and related ops
- add tests
- move 2D objects into a dedicated geometry submodules

## Scope

## Type of change

- [x] New feature(s)

### DIFF
--- a/honeycomb-core/src/geometry/dim3/tests.rs
+++ b/honeycomb-core/src/geometry/dim3/tests.rs
@@ -1,0 +1,195 @@
+// ------ IMPORTS
+
+use super::super::{CoordsFloat, Vector3, Vertex3};
+
+// ------ CONTENT
+
+// --- common
+// scalar equality function
+// ref: https://floating-point-gui.de/errors/comparison/
+fn almost_equal<T: CoordsFloat>(a: T, b: T) -> bool {
+    let abs_diff = (a - b).abs();
+    let abs_sum = a.abs() + b.abs();
+    if a == b {
+        // early return
+        true
+    } else if a.is_zero() | b.is_zero() | (abs_sum < T::min_positive_value()) {
+        // close to zero
+        abs_diff < (T::epsilon() * T::min_positive_value())
+    } else {
+        // regular case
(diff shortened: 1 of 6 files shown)


honeycomb-core/src/geometry/dim3/tests.rs
```diff
@@ -0,0 +1,220 @@
...
+// scalar equality function
+// ref: https://floating-point-gui.de/errors/comparison/
+fn almost_equal<T: CoordsFloat>(a: T, b: T) -> bool {
+    let abs_diff = (a - b).abs();
+    let abs_sum = a.abs() + b.abs();
+    if a == b {
+        // early return
+        true
+    } else if a.is_zero() | b.is_zero() | (abs_sum < T::min_positive_value()) {
+        // close to zero
```
cedricchevalier19, line 17: Is `T::min_positive_value()` the smallest positive *normal* value?

I do not quite follow this path. Why is using a non-zero *denormal* value a problem?

honeycomb-core/src/geometry/dim3/tests.rs
```diff
@@ -0,0 +1,220 @@
+// ------ IMPORTS
+
+use super::super::{CoordsFloat, Vector3, Vertex3};
+
+// ------ CONTENT
+
+// --- common
+// scalar equality function
+// ref: https://floating-point-gui.de/errors/comparison/
```
cedricchevalier19, line 9: I am not totally convinced by the scaling by |A|+|B|.

Can we use a specialized crate like https://docs.rs/float-cmp/latest/float_cmp/ for this kind of stuff?

cedricchevalier19 (approved):
I think the floating point comparisons should be questioned elsewhere.

Do you think that a conversion from `Vertex2` to `Vertex3` should be useful?

imrn99:
I'll address all issues regarding comparison in another PR since it will also apply to `Vertex2`. Other stuff I'll add to this PR.

> Do you think that a conversion from `Vertex2` to `Vertex3` should be useful?

Yes, I'll add it. It will be required if we implement CMap2 to CMap3 conversion later.

PR #199, opened 2024-10-17
feat(core): implement parallel variants of attribute storage traits & structs

imrn99:
- add `ParUnknownAttributeStorage`, with update method sigs
- add `ParAttributeStorage`, with update method sigs
- implement these for a new struct, `PAttrSparseVec`

### DIFF
--- a/honeycomb-core/src/attributes/collections.rs
+++ b/honeycomb-core/src/attributes/collections.rs
@@ -293,6 +293,21 @@ impl<T: AttributeBind + AttributeUpdate + Clone> AttrCompactVec<T> {
 
 // --- parallel items
 
+/// Custom storage structure for attributes
+///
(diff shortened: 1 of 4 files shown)


cedricchevalier19 (commented):
Can you test the usefulness of the lock?

Can you comment on why you chose this `Vec<RwLock<>>` solution.
A `RwLock<Vec<>>` solution is also possible, can you comment on the differences?

imrn99:
> Can you test the usefulness of the lock?

I'm not sure if I can test it easily, but I can demonstrate its usefulness

> Can you comment on why you chose this `Vec<RwLock<>>` solution. A `RwLock<Vec<>>` solution is also possible, can you comment on the differences?

Given the conversation in #199, I'll close this PR and use the same approach here (change the original struct). I'll do everything in #201, with a dedicated comment per change in the PR

PR #200, opened 2024-10-17
refactor!(core): create a `common` module in core crate

imrn99:
- create the new module
- move `DartIdentifier`, `NULL_DART_ID` to it
- move `CMapBuilder` a all related items to it


imrn99:
postponing this due to change in approach. I'll probably do something similar when 3D is implemetned?

PR #201, opened 2024-10-17
refactor!(core): change underlying implementation of `CMap2` to add sync mechanisms

imrn99:
- remove the `Clone` trait impl when feature `utils` was enabled (unused after some benchmark deletion)
- comment out `AttrCompactVec` for now
- replace internal beta storage type `u32` by its atomic equivalent `AtomicU32`
- replace unused darts `BTreeSet`by `Vec<AtomicBool>`
- change `AttrSparseVec` internal storage type `Option<A>` by `Atomic<Option<A>>` using the `atomic` crate
- Update signatures from `CMap2`, storage traits methods, storage manager, to be non-mutable
- add `Send` & `Sync` implementations for:
    - `AttrSparseVec`, `AttrStorageManager`
    - `CMap2`
    - `VertexCollection`, `EdgeCollection`, `FaceCollection`
    - `Vertex2`, `Vector2`


imrn99:
## `Clone` impl / `AttrCompactVec` removal

this is a convenience change. both weren't used currently and require significant changes to make them work in parallel. this can be done in a separate PR

imrn99:
## Internal beta storage

This is the storage structure of beta functions:

betas

values were stored using `u32` (through an alias named `DartIdentifier`). the internal type was changed to `AtomicU32`, while method & function signatures still take `DartIdentifier`s as arguments.

atomic load/write is handled internally, with no change to the public API (aside from mutability requirement). all operations on these are using a relaxed ordering, but this is subject to change.

the external vector (red) is not wrapped in a synchronization layer as we consider that expending, reallocating, etc. should be done before heavily-parallel computation kernels. if, at any point, we need to do this, the mechanism used should probably be a `RwLock`.

imrn99:
## unused darts tracking

reminder: dart removal means (a) set its beta values to `NULL_DART_ID` (`== 0`), and (b) mark it as unused

if we take the [last step of the clipping routine](https://lihpc-computational-geometry.github.io/honeycomb/honeycomb_kernels/grisubal/index.html#post-processing-clip) of the `grisubal` algorithm as an example, the problem with the original data structure is obvious: after a coloring algorithm, we mass-delete darts that are not part of final structure.

if we were to parallelize this mass-deletion step, many all threads would be contending over access to a single `BTreeSet`, to insert the dart being deleted. by replacing the set with a vector of atomics, we can ensure that there won't be contention over this operation, assuming that two threads do not attempt to delete the same dart (which is an incorrect operation anyway)

the vector itself is not wrapped in a sync layer for the same reason mentionned in the beta function storage:

> the external vector (red) is not wrapped in a synchronization layer as we consider that expending, reallocating, etc. should be done before heavily-parallel computation kernels. if, at any point, we need to do this, the mechanism used should probably be a `RwLock`.

imrn99:
## vertex / attribute storage

the issue is illustrated here:

attributesync

these storages require specific sync. because there is an intermediate state where not all of the involved data has been attributed its final values, we need to ensure nothing can happen in that time frame for operation correctness.

a first idea was to wrap each entry in `RwLock`s. while the implementation worked, it was discarded for something simpler that *should* work (needs testing using something like [loom](https://github.com/tokio-rs/loom).

the solution is to wrap attribute values in atomics, and use `Acquire`/`Release`/`AcqRel` ordering to ensure there's no problematic instruction re-ordering. the idea came from @dssgabriel, and we speculated that these orderings are sufficient for the operation described.

honeycomb-core/src/attributes/tests.rs
```diff
@@ -525,8 +528,9 @@ fn manager_vec_insert_already_existing() {
 #[test]
 fn manager_vec_remove() {
     generate_manager!(manager);
-    generate_compact!(storage);
-    assert_eq!(storage.remove(3), Some(Temperature::from(279.0)));
+    // what was this ?
```
cedricchevalier19, line 531: Good question!

honeycomb-core/src/cmap/dim2/basic_ops.rs
```diff
@@ -21,7 +21,7 @@ use crate::{
 };
 
 use std::collections::BTreeSet;
-
+use std::sync::atomic::{AtomicBool, Ordering};
```
cedricchevalier19, line 24: Why do not use transactions here too?

cedricchevalier19 (commented):
The mix between Atomic and Transactions is hard to follow. Can you explain what has led to these choices?

honeycomb-core/src/cmap/dim2/basic_ops.rs
imrn99: Ok, there's two things:

- for unused dart storage, `AtomicBool` is enough because there's no situation / operation where the validity of a dart is a critical information. This could change if we implement algorithms that delete part of the mesh sporadically, but even then it doesn't matter: What is important is that all beta function of the deleted dart are set to 0, not that it is marked. marks matter when fetching all cells after (or before) an algorithm is executed.

- for beta values: these should probably be transactional too. the biggest consequence would be that some method (e.g. attribute collection methods) need to take a transaction as argument as the `atomically` section provided byt he stm crate should not be nested

imrn99:
I rewrote a few examples on a white board and found cases where atomics on beta values were satisfying. given this, I changed the storage type to `TVar<DartIdentifier>`. I reimplemented methods as needed to avoid nesting transaction in higher level ops (e.g. sew). this was made easy by the `stm` crate as nested transaction result in a crash at runtime.

these changes significantly enlarge the PR, but some cleanup is possible. as of now, I duplicated most original methods to add variants taking a `Transaction` as argument. it might be better to replace the original methods, although it will absolutely break the API

cedricchevalier19:
Ok, I will take my time to understand the changes.

imrn99:
following a conversation with @cedricchevalier19, this is the course of action  we'll follow:
- switch `unused_darts` storage to transactional variables too, in order to fully adopt the model and its guarantees
- merge this PR
- write tests and merge them in a separate PR
- cleanup in a separate PR

PR #203, opened 2024-10-21
refactor(kernels::grisubal): use prealloc methods for edge splitting

imrn99:
preallocation occurs in step 2 and step 4. note that:
- the computation done for preallocs are probably not optimal
- step 2 can be further modified to (a) make it more efficient, and (b) make step 3 independent from it

### DIFF
--- a/honeycomb-kernels/src/grisubal/kernel.rs
+++ b/honeycomb-kernels/src/grisubal/kernel.rs
@@ -510,16 +510,19 @@ pub(super) fn insert_intersections<T: CoordsFloat>(
     // b.
     // FIXME: minimize allocs & redundant operations
     // prealloc all darts needed
-    let ns_darts: Vec<_> = edge_intersec.values().map(|vs| 2 * vs.len()).collect();
-    let n_tot: usize = ns_darts.iter().sum();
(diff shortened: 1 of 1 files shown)


honeycomb-kernels/src/grisubal/kernel.rs
```diff
@@ -509,11 +508,39 @@ pub(super) fn insert_intersections<T: CoordsFloat>(
         });
 
     // b.
-    for (edge_id, vs) in &mut edge_intersec {
+    // FIXME: minimize allocs & redundant operations
```
cedricchevalier19, line 511: Would you happen to have an idea of the overestimate?

cedricchevalier19 (approved):

honeycomb-kernels/src/grisubal/kernel.rs
imrn99: I'm not 100% sure, I added the comment because I didn't think a lot about the impl and just wrote it as it came to me. I hope there is a better way to implement this because the number of iterators & `.collect` calls seems atrocious. 

The good news is that there is no change of performance with this impl in sequential, so it's at least as good as the previous impl while enabling parallelization of the next loop.

PR #205, opened 2024-10-24
refactor(kernels::grisubal): add intermediate compute before step 2

imrn99:
- convert step 2 function to only cover vertex insertions
- move other components out of it to (a) clearup code, and (b) make step 3 independent from step 2. these components are:
    - grouping intersection per edge intersected
    - pre-allocating darts used for step 2
    - computing each intersection's corresponding new dart

### DIFF
--- a/honeycomb-kernels/src/grisubal/kernel.rs
+++ b/honeycomb-kernels/src/grisubal/kernel.rs
@@ -170,6 +170,14 @@ pub fn build_mesh<T: CoordsFloat>(
     cmap
 }
 
+// --- type aliases for clarity
+
+pub type Segments = HashMap<GeometryVertex, GeometryVertex>;
+
+pub type IntersectionsPerEdge<T> = HashMap<EdgeIdentifier, Vec<(usize, T, DartIdentifier)>>;
+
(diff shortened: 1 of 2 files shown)


honeycomb-kernels/src/grisubal/kernel.rs
```diff
@@ -531,39 +544,51 @@ pub(super) fn insert_intersections<T: CoordsFloat>(
...
+    for ((edge_id, vs), new_darts) in edge_intersec.iter().zip(dart_slices.iter()) {
         // order should be consistent between collection because of the sort_by call
-        let mut dart_id = cmap.beta::<1>(*edge_id as DartIdentifier);
-        // chaining this directly avoids an additional `.collect()`
-        for (id, _, old_dart_id) in vs {
-            // c.
-            // reajust according to intersection side
+        let hl = new_darts.len() / 2; // half-length; also equal to n_intermediate
+        let fh = &new_darts[..hl];
+        let sh = &new_darts[hl..];
```
cedricchevalier19, line 575: What fh and sh stand for?

cedricchevalier19 (approved):

PR #207, opened 2024-10-24
feat(benches): add a criterion bench for triangulation fucntions

imrn99:
- add a single benchmark group, where exec time of both `fan_cell` and `ear_clip` is measured
- add a sample mesh file for the bench. the mesh is made up of both tris and quads


imrn99:
ok maybe the 21K lines vtk file is a bad idea

cedricchevalier19 (approved):

PR #209, opened 2024-10-28
refactor(core): imrove grid building routines

imrn99:
1. remove a useless if / wrong range in the second construction loop
2. fuse and flatten both construction loops, removing redundant computations at the same time
3. replace `link` usages with `set_beta` usages
4. set the last assertion to debug only (it can be removed entirely I think).

performance changes for a 512x512 grid (exec time in ms):

| Grid type | Original | new fetch_<CELLS> | Optim 1, 2, 3 | Optim 4 |
| --------- | --------:| -----------------:| -------------:| -------:|
| Square    | `157.11` |           `89.83` |       `92.98` | `24.54` |
| Triangle  | `246.48` |          `120.16` |      `121.79` | `35.54` |

### DIFF
--- a/honeycomb-core/src/cmap/builder/grid/building_routines.rs
+++ b/honeycomb-core/src/cmap/builder/grid/building_routines.rs
@@ -36,12 +36,6 @@ pub fn build_2d_grid<T: CoordsFloat>(
                 let up_neighbor = d1 + (4 * n_square_x) as DartIdentifier;
                 map.two_link(d3, up_neighbor);
             }
-        });
-    });
-
-    // then cells
-    (0..n_square_y).for_each(|y_idx| {
-        (0..n_square_x).for_each(|x_idx| {
             // update the associated 0-cell
             let base_dart = (1 + 4 * x_idx + n_square_x * 4 * y_idx) as DartIdentifier;
             let vertex_id = map.vertex_id(base_dart);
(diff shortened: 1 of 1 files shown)


honeycomb-core/src/cmap/builder/grid/building_routines.rs
```diff
@@ -17,88 +17,93 @@ pub fn build_2d_grid<T: CoordsFloat>(
     let mut map: CMap2<T> =
         CMap2::new_with_undefined_attributes(4 * n_square_x * n_square_y, manager);
 
-    // first, topology
-    (0..n_square_y).for_each(|y_idx| {
-        (0..n_square_x).for_each(|x_idx| {
+    (0..n_square_y)
+        .flat_map(|y_idx| (0..n_square_x).map(move |x_idx| (y_idx, x_idx)))
```
cedricchevalier19, line 21: Comment on why flatten the iterations.

honeycomb-core/src/cmap/builder/grid/building_routines.rs
```diff
@@ -17,88 +17,93 @@ pub fn build_2d_grid<T: CoordsFloat>(
...
+                            T::from(y_idx + 1).unwrap() * len_per_y,
+                        ),
+                );
             }
         });
-    });
 
     // and then build faces
-    assert_eq!(map.fetch_faces().identifiers.len(), n_square_x * n_square_y);
+    debug_assert_eq!(map.fetch_faces().identifiers.len(), n_square_x * n_square_y);
```
cedricchevalier19, line 106: Leave a comment stating why it can impact the performances.

cedricchevalier19 (approved):
I was just wondering if splitting the loop into (the interior domain and then the boundary (last row & last column)) can not make the code easier to read (and perhaps more efficient without a conditional in the middle).

honeycomb-core/src/cmap/builder/grid/building_routines.rs
imrn99: done in 5d61aa9

honeycomb-core/src/cmap/builder/grid/building_routines.rs
imrn99: done in 5d61aa9

imrn99:
> I was just wondering if splitting the loop into (the interior domain and then the boundary (last row & last column)) can not make the code easier to read (and perhaps more efficient without a conditional in the middle).

I managed to reorganize the code by splitting like you suggested. I also extracted some common functions which should help with readability

PR #214, opened 2024-10-30
test(core): add coverage over new parallel implementations

imrn99:
this adds coverage over additions of #201 

it also fixes a problem highlighted by the new tests; the `two_unsew` methods were not correctly dealing with vertex attributes


imrn99:
the windows CI has decided it won't run :+1:

cedricchevalier19 (approved):

PR #216, opened 2024-10-31
refactor(kernels::grisubal): clean-up module code

imrn99:
- rename the `kernel` module to `routines`, split its original content into file
- turn step 1.5 into step2, offset all subsequent steps
- add pre-processing routines to the `routines` module
- delete the `grid` module, move its content to `model`
- clean the timers implementation to hide the `#[cfg(feature = "profiling")]` attributes
- remove the `build_mesh` function to call directly step functions in the main `grisubal` function

I will update the doc of the algorithm in a separate PR since I intend to change more things in the doc


honeycomb-kernels/src/grisubal/model.rs
```diff
@@ -6,38 +6,39 @@
...
-    None,
+/// Cells `(X, Y)` take value in range `(0, 0)` to `(N, M)`,
+/// from left to right (X), from bottom to top (Y).
+#[derive(PartialEq, Clone, Copy)]
+pub struct GridCellId(pub usize, pub usize);
+
+impl GridCellId {
+    /// Compute the [Manhattan distance](https://en.wikipedia.org/wiki/Taxicab_geometry) between
+    /// two cells.
+    pub fn man_dist(lhs: &Self, rhs: &Self) -> usize {
```
cedricchevalier19, line 31: This name is confusing. `L1` or `distance1` or something like this is more readable if you want to keep it short.

honeycomb-kernels/src/grisubal/model.rs
```diff
@@ -6,38 +6,39 @@
...
+        lhs.0.abs_diff(rhs.0) + lhs.1.abs_diff(rhs.1)
+    }
+
+    #[allow(clippy::cast_possible_wrap)]
+    pub fn diff(lhs: &Self, rhs: &Self) -> (isize, isize) {
```
cedricchevalier19, line 36: Same here. Is `diff` for difference, like subtraction?

cedricchevalier19 (commented):
I am not convinced by putting the functions into step"I" files.

I think keeping a comment in the file and mod.rs that it is used in step "I" is more useful. And, you can name the files according to their contents.

honeycomb-kernels/src/grisubal/model.rs
imrn99: renamed to `l1_dist`

honeycomb-kernels/src/grisubal/model.rs
imrn99: renamed to `offset`. It is not exactly a substraction either since there is a notion of direction (offset from lhs to rhs), so I added some doc.

imrn99:
I renamed all module files to describe the step & added comments in the `routines` module for order.

cedricchevalier19 (approved):
LGTM

PR #220, opened 2024-11-08
refactor!: replace identifier type aliases with strong-typed structures

imrn99:
- Betas & unused darts use dedicated structs for readability
- All identifier type aliases are swapped out for:
    - strongly typed struct (e.g. `DartId`)
    - an inner type alias (e.g. `DartIdType`)
- All methods and function that were taking type aliases as args now take the storngly typed structure instead
- cell id structs all implement `From` (both ways) for `DartId`, but do not implment `From` between each others
- fix an issue in `atomically_two_sew`. I'll change these methods soon anyway

a lot of these changes are noise from examples & tests that used direct integer values as args


imrn99:
putting this on hold until I figure out why the exec time is 20% higher on benchmarks

PR #224, opened 2024-11-13
refactor!(core): rework `sew` methods to allow for more control

imrn99:
I split the original `link_and_sew.rs` file into two modules: `links` and `sews`, each having submodules for dim1 and dim2.

there are now 3 and 2 variants for (un)sew and (un)link methods respectively. Here are two tables summarizing their differences (for `sew` and `link`, complementary methods follow the same pattern):

| sew variant | description |
|-----------------| --- |
| `try_<DIM>_sew` | defensive impl, only succeding if attributes are correctly merged & the transaction is validated |
| `<DIM>_sew` | regular impl, which uses attribute fallback policies and will fail only if the transaction fails |
| `force_<DIM>_sew` | convenience impl, which follows the same logic as the regular impl, but using a transaction internally to retry until success. useful when precise control isn't needed (e.g. in sequential impls) |

| link variant | description |
|--------------| --- |
| `<DIM>_link` | regular impl, which will fail if the transaction fails |
| `force_<DIM>_sew` | convenience impl, which follows the same logic as the regular impl, but using a transaction internally to retry until success. useful when precise control isn't needed (e.g. in sequential impls) |

TODO: implement the `try_` variants. I want to do it in a separate PR because I will modify the attribute storage traits to follow a similar pattern. I'm guessing do it here would double the size of the PR, and unnecessarily mix unrelated changes


imrn99:
> An a priori comment: I am not convinced by using `Result` for the transactions. Do you have a strong argument for using this instead of `Option` ?

I'm not sure to exactly which `Result` you refer to. The `stm` crate originally uses results to pass value through (see [`StmResult`](https://docs.rs/stm-core/0.4.0/stm_core/type.StmResult.html)), so I aligned the API with this.

In our case, we could switch the `Result` to an `Option` for all instances of `Result<(), _>`, if that's what you mean

cedricchevalier19:
No, do not change, I was wondering the reason. But mimicking the stm crate is a pretty valid one.

cedricchevalier19 (approved):
Looks Ok.

PR #238, opened 2024-11-25
refactor!(core): add new variants to `OrbitPolicy`

### LINKED ISSUE #235: [Improvement]: add more `OrbitPolicy` variants
### Scope

honeycomb-core (main structures)

### Side effects

unsure

### Type of change

new feature

### Affected item(s)

- `OrbitPolicy`
- `Orbit2` computation routines

### Changes / proposals

Because we use the beta 0 function in our implementation, some i-cells can be computed using two different orbits. For example a face can be computed using only beta 1, or both beta 1 and beta 0. Both compute methods are often useful, so we want to support both.

- add new variants to `OrbitPolicy` such as `FaceLinear` (name TBD)
- implement routines for new variants (trivial)

imrn99:
### Description

**Associated issue**: closes #235 

**Content description**:
- add two new variants `OrbitPolicy::VertexLinear` & `OrbitPolicy::FaceLinear`
- handle their computation in `Orbit2`
- change `OrbitPolicy::Face` to use the beta 0 & beta 1 functions (it previously only used beta 1, which is what `OrbitPolicy::FaceLinear` does now)

### Additional information

- [x] Breaking change

### DIFF
--- a/honeycomb-core/src/attributes/manager.rs
+++ b/honeycomb-core/src/attributes/manager.rs
@@ -21,9 +21,11 @@ use std::{any::TypeId, collections::HashMap};
 macro_rules! get_storage {
     ($slf: ident, $id: ident) => {
         let probably_storage = match A::BIND_POLICY {
-            OrbitPolicy::Vertex => $slf.vertices.get(&TypeId::of::<A>()),
+            OrbitPolicy::Vertex | OrbitPolicy::VertexLinear => {
+                $slf.vertices.get(&TypeId::of::<A>())
+            }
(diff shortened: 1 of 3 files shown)


cedricchevalier19 (approved):
What is the use case for the new policies?

imrn99:
> What is the use case for the new policies?

Taking faces as an example:
- `FaceLinear`: If I need to compute a value from a face's edges, I'll try to build a sliding window iterator over an (ordered) list of its vertices. 
- `Face`: If I'm searching for a vertex to star a polygon from, I probably want to search in both direction for efficiency. see very rough drawing below, though here it's an obvious edge case

Untitled

PR #240, opened 2024-11-27
feat(benches): add binaries to bench scaling of non-conflicting transactions

imrn99:
### Description

**Scope**:  `benches`

**Type of change**: new feature

**Content description**: 

add two binaries, `shift` and `shift-nc`. both programs generate a grid and iterate on inner vertices, computing the average of their neighbors and taking it as new value. this is repeated for N rounds.

the main loop is parallelized using `rayon`, the easiest way to control execution resource is to use `taskset`.

`shift-nc` differs from `shift` by splitting the vertices into to independent batches before processing those. the intended effect is to decouple iterations that could conflict with each other, resulting in no transaction falures and retries.

### Additional information

- [x] New dependency: `rayon`

### Possible follow-up

- larger scale tests
- add parallelism to predefined kernels


benches/src/shift.rs
```diff
@@ -0,0 +1,66 @@
+use rayon::prelude::*;
+
+use honeycomb::core::stm::atomically;
+use honeycomb::prelude::{
+    CMap2, CMapBuilder, DartIdType, Orbit2, OrbitPolicy, Vertex2, VertexIdType, NULL_DART_ID,
+};
+
```
cedricchevalier19, line 7: Add a small description of the program

benches/src/shift_no_conflict.rs
```diff
@@ -0,0 +1,84 @@
+use rayon::prelude::*;
+
+use honeycomb::core::stm::atomically;
+use honeycomb::prelude::{
+    CMap2, CMapBuilder, DartIdType, Orbit2, OrbitPolicy, Vertex2, VertexIdType, NULL_DART_ID,
+};
+
```
cedricchevalier19, line 7: Add a small description of the program

cedricchevalier19 (commented):
The two new programs are benchmarks, but what is the performance metric?

imrn99:
> The two new programs are benchmarks, but what is the performance metric?

The main metric of interest is execution time, to evaluate how well our implementation scales. Both weak and strong scaling can be estimated by controlling grid size and # of thread used.

We don't necessarily care about flat exec time since we don't have anything to compare against, and I'm not even sure the algorithm makes too much sense meshing-wise.

imrn99:
As an indication, I'm reaching quasi ideal speedup for a 512*512 map, from 1 to 12 threads (from 92s down to 12s, given that map creation is not parallelized)

cedricchevalier19 (approved):

PR #241, opened 2024-11-29
refactor(core): rewrite grid building routines

imrn99:
### Description

**Scope**:  core

**Type of change**: refactor of internals

**Content description**:

The new routines compute separately the value of the beta functions and manually store them for each dart (instead of using sews and links). Vertices are added after beta functions are fully defined.

There is a slight improvement in performance (5-10%), but this is not the interesting change. With the new routines, there won't be any conflict between transactions when parallelized. every write operation is containerized, and does not affect data read by another transaction.


imrn99:
CI fails due to new Rust version, I'll fix it in another PR

cedricchevalier19 (approved):
Perhaps you can add a comment or a link to the documentation about how the ordering works.

PR #245, opened 2024-12-03
feat(kernel): implement DSATUR coloring algorithm for vertices

### LINKED ISSUE #236: [Improvement]: implement a coloring algorithm
### Scope

honeycomb-kernels (algorithms)

### Side effects

no

### Type of change

new feature

### Affected item(s)

- new module in `honeycomb-kernels`

### Changes / proposals

Some form of coloring algorithm could be implemented as they both:

1. present some characteristics that we haven't explored yet (read-heavy, possibly without edition if colors are not included in the map)
2. are useful to other algorithms (e.g. identify independent workloads to parallelize)

vertex coloring should be first since I already identified useful cases (2)

imrn99:
### Description

**Associated issue**: closes #236 

**Content description**:

- add `add_attribute_storage` method to `CMap2` for convenience
- implement the DSATUR coloring algorithm. ref: https://fr.wikipedia.org/wiki/DSATUR. FR page has better step descriptions 

**Follow-up**

Aside from using it, I think the algorithm implementation can be modified to handle coloring on all types of I-cells, using a trait and the `enum_dispatch` crate to avoid dyn object lookup costs.

### DIFF
--- a/honeycomb-kernels/src/coloring/dsatur.rs
+++ b/honeycomb-kernels/src/coloring/dsatur.rs
@@ -6,6 +6,7 @@ use honeycomb_core::{
     stm::atomically,
 };
 use itertools::Itertools;
+use rayon::prelude::*;
 
 use super::Color;
 
@@ -26,7 +27,7 @@ pub fn color<T: CoordsFloat>(cmap: &mut CMap2<T>) -> u8 {
     let nodes: Vec<(VertexIdType, Vec<VertexIdType>)> = cmap
         .fetch_vertices()
         .identifiers
-        .into_iter()
+        .into_par_iter()
         .map(|v| {
             (
                 v,
@@ -86,11 +87,8 @@ pub fn color<T: CoordsFloat>(cmap: &mut CMap2<T>) -> u8 {
         c = 0;
     }
 
-    atomically(|trans| {
-        for (&v, &c) in &colors {
-            cmap.write_attribute(trans, v, c)?;
-        }
-        Ok(())
+    colors.into_par_iter().for_each(|(v, c)| {
+        atomically(|trans| cmap.write_attribute(trans, v, c));
     });
 
     cmax
--- a/honeycomb-kernels/src/coloring/dsatur.rs
+++ b/honeycomb-kernels/src/coloring/dsatur.rs
@@ -1,22 +1,75 @@
-use std::collections::HashMap;
+use std::{cmp::Ordering, collections::HashMap};
 
 use honeycomb_core::{
     cmap::{CMap2, DartIdType, Orbit2, OrbitPolicy, VertexIdType, NULL_DART_ID},
     prelude::CoordsFloat,
 };
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 struct Color(u8);
 
+/// DSATUR algorithm implementation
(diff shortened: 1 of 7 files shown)


honeycomb-kernels/src/coloring/dsatur.rs
```diff
@@ -0,0 +1,95 @@
...
+/// DSATUR algorithm implementation
+///
+/// This algorithm is a coloring algorithm similar to the greedy coloring algorithm.
+///
+/// [Reference](https://en.wikipedia.org/wiki/DSatur).
+///
+/// # Return
+///
+/// The function will return the maximal used color, i.e. the `n_color_used - 1`.
+pub fn color<T: CoordsFloat>(cmap: &mut CMap2<T>) -> u8 {
```
cedricchevalier19, line 23: Not sure if it is a good name. Perhaps the algorithm name will be better. Add to the description it is a distance-1 coloring of the vertices.

honeycomb-kernels/src/coloring/dsatur.rs
```diff
@@ -0,0 +1,95 @@
...
+    cmap.add_attribute_storage::<Color>();
+
+    // build graph data as a collection of (Vertex, Vec<Neighbors>)
```
cedricchevalier19, line 26: Could this part be helpful elsewhere?
You can add a way to create a graph from a CMap. And use things like Coupe later ;).

honeycomb-kernels/src/coloring/dsatur.rs
```diff
@@ -0,0 +1,95 @@
...
+        }
+        cmax = cmax.max(c);
+        colors.insert(*v, Color(c));
+
+        // find next candidate that is:
+        //
+        // - not colored
+        // - of highest saturation
+        //   - if there are multiple, take the one of highest degree
+        crt_node = nodes
```
cedricchevalier19, line 75: Can you comment on the complexity?

cedricchevalier19 (commented):
I am not fully convinced by the implementation.

honeycomb-kernels/src/coloring/dsatur.rs
imrn99: I re-export the function as `color_dsatur` in the `coloring` module. I can switch it to `dsatur` if you want

honeycomb-kernels/src/coloring/dsatur.rs
imrn99: I'll add it to the doc, but it's O(n²)

honeycomb-kernels/src/coloring/dsatur.rs
imrn99: > Could this part be helpful elsewhere? You can add a way to create a graph from a CMap. 

Yeah, I think I can also make it generic enough to handle all possible adjacency types (e.g.  here we're using vertex-vertex, but we could fetch incident faces for another usage). 

> And use things like Coupe later ;).

I actually wanted to try and use it, but I think all coloring algorithms of coupe are focused on balancing weights, rather than making independent subsets. Is that right?

imrn99:
I'm closing this since:
- this isn't needed atm
- integrating with `coupe` seems much more interesting

PR #247, opened 2024-12-05
refactor!(core): simplify transaction API

imrn99:
### Description

**Scope**:  core 

**Type of change**: refactor 

**Content description**:

- Remove the regular variants and replace it with the `try_` variant. The latter has been renamed accordingly
- The behavior of the `force_` variant has not changed, but it is not a wrapper around the regular variant anymore: It ignores attribute-related failure and retry the transaction until complete  

### Additional information

- [x] Breaking change

### DIFF
--- a/honeycomb-core/src/cmap/error.rs
+++ b/honeycomb-core/src/cmap/error.rs
@@ -20,6 +20,9 @@ pub enum CMapError {
     /// Geometry check failed.
     #[error("operation incompatible with map geometry: {0}")]
     IncorrectGeometry(&'static str),
+    /// No attribute value associated to specified ID.
+    #[error("missing attribute: {0}")]
+    MissingAttribute(&'static str),
     /// Accessed attribute isn't in the map storage.
     #[error("unknown attribute: {0}")]
     UnknownAttribute(&'static str),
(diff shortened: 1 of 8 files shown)


imrn99:
I have opted out of replacing all `Result<Option<A>>` return types for ergonomy reasons. If I replace these with a dedicated error type, calling the `?` won't have the same effects, and it complicates transaction behavior control.

honeycomb-core/src/cmap/error.rs
```diff
@@ -30,3 +30,16 @@ impl From<StmError> for CMapError {
         Self::FailedTransaction(value)
     }
 }
+
+// by default, a map error inside of a transaction will result in a retry
+impl From<CMapError> for StmError {
```
imrn99, line 35: FWIW, I think this is my main issue with this PR

cedricchevalier19:
If we can discuss it tomorrow that will help me a lot for the review.

imrn99:
I'm probably going to close this PR in favor of a new one which will include changes from #253. The idea is still to delete a variant, but this time I won't add the automatic error conversion which annoys me here.

imrn99:
replaced by #261

PR #249, opened 2024-12-09
feat(core): add minimal `CMap3` implementation

imrn99:
### Description

**Scope**:  core

**Type of change**: feat

**Content description**:

Add a minimal implementation of the 3D combinatorial maps. The implementation only includes basic operations, no utilities / predefined routines / kernels / ...

### Necessary follow-up

implement all the side stuff this PR doesn't include, most notably:
- IO
- integration to the builder 
- grid building?
- render?

also:
- remove dead code of the attribute module
- push further for coverage of base structs?
- improve orientation check for 3 sew (check all dart pais instead of just args)


imrn99:
to make the implementation cleaner, there are a few changes I'm gonna do in other PRs, which I'll fetch back in this branch:
- [ ] remove the `CMap2` ref from i-cell collection structs, maybe remove these structs entirely & use lifetime'd iterators instead
- [ ] validate the sew API (see #247)
- [ ] figure out what code can be extracted from dim2 orbits / I'll probably re-evaluate the impl later to hardcode more things

imrn99:
TODO:
- [ ] implement sews according to changes made in #253 and TBD
- [ ] add documentation

imrn99:
common methods of the structure are tested. coverage is about the same as `CMap2`, but new methods added to the manager brind down the global coverage. a lot of these methods are unused after changes of #261, I intend to delete these in a separate PR.

the clippy CI doesn't pass due to a new lint, which I'd argue is terrible for readability in our case. I'll fix it (or make it an exception) in another PR

cedricchevalier19 (approved):
LGTM.

And thanks for explaining the coverage.

Perhaps we can spend some time after to identify if the code can be shared between all the l-links. But currently, it is pretty readable.

PR #253, opened 2024-12-12
refactor!(core): make dimension an argument of sew/link methods

imrn99:
### Description

**Scope**:  core with lots of side effects

**Type of change**: refactor

**Content description**:

- added new sew and link methods taking dimension as a const argument
- old sew and links methods are now private 

With assertions and match on a const generic, the indirection layer has no cost (with optimization enabled). Small example with compiler explorer:

https://godbolt.org/z/857KTob9T

### Additional information

- [x] Breaking change

### Change example

The new `link` is implemented as:

```rust
    pub fn link<const I: u8>(
        &self,
        trans: &mut Transaction,
        lhs_dart_id: DartIdType,
        rhs_dart_id: DartIdType,
    ) -> StmResult<()> {
        // these assertions + match on a const are optimized away
        assert!(I < 3);
        assert_ne!(I, 0);
        match I {
            1 => self.one_link(trans, lhs_dart_id, rhs_dart_id),
            2 => self.two_link(trans, lhs_dart_id, rhs_dart_id),
            _ => unreachable!(),
        }
    }
```


cedricchevalier19 (approved):
I like this syntax

PR #256, opened 2024-12-13
feat(core): implement 3D grid generation internal routines

imrn99:
### Description

**Scope**: core

**Type of change**: feat

**Content description**:

- rename `GridDescriptor::origin` to `GridDescriptor::origin_2d` and add `GridDescriptor::origin_3d`
- add grid descriptor 3D parsing routine
- add internal routines for 3D grid generation
    - `generate_hex_beta_values` for beta function values
    - `generate_hex_offset` for vertex coordinate values

The vertex computing function uses a different logic than the its 2D counter part. Essentially, a value is computed for every single dart, but the only instances used are the ones corresponding to an actual vertex ID. This emsures value generation won't ever be the bottleneck as generating these values can be trivially done in parallel / on gpu (see this [repo](https://github.com/imrn99/poc-honeycomb-gpu))

### Follow up

- add a `build_3d` method to `CMapBuilder`
- replace the `From<PathBuf>` impl of `CMapBuilder`
- add tests


imrn99:
TODO:
- [ ] add vertex insertion
- [ ] add another drawing of the expected IDs?
- [ ] wire the routines to the builder
- [ ] tests

honeycomb-core/src/cmap/builder/grid.rs
```diff
@@ -17,7 +17,8 @@ use crate::geometry::{CoordsFloat, Vector2, Vertex2};
 /// - `T: CoordsFloat` -- Generic FP type that will be used by the map's vertices.
 #[derive(Default, Clone)]
 pub struct GridDescriptor<T: CoordsFloat> {
-    pub(crate) origin: Vertex2<T>,
+    pub(crate) origin2: Vertex2<T>,
```
cedricchevalier19, line 20: Why does it not have dimensionality as a template parameter? Or can we have a `LinearGridDescriptor` and an `AffineGridDescriptor` that deals with the offset?

Below, you should rename `split_quads` as it is likely to be something else in 3d.

honeycomb-core/src/cmap/builder/grid.rs
```diff
@@ -87,8 +88,15 @@ impl<T: CoordsFloat> GridDescriptor<T> {
 
     /// Set origin (most bottom-left vertex) of the grid
     #[must_use = "unused builder object"]
-    pub fn origin(mut self, origin: Vertex2<T>) -> Self {
-        self.origin = origin;
+    pub fn origin_2d(mut self, origin: Vertex2<T>) -> Self {
```
cedricchevalier19, line 91: Likewise, I am not convinced that we need to have different method names. Or do you have a case that uses both `origin2` and `origin3`?

honeycomb-core/src/cmap/builder/grid.rs
```diff
@@ -379,3 +453,286 @@ fn generate_tris_beta_values(n_x: usize, n_y: usize) -> impl Iterator<Item = [Da
...
+
+// ------ 3D
+
+#[allow(unused)]
+/// Internal grid-building routine
+#[allow(clippy::too_many_lines)]
+pub fn build_3d_grid<T: CoordsFloat>(
+    origin: Vertex3<T>,
+    ns: [usize; 3],
+    lens: [T; 3],
```
cedricchevalier19, line 465: I am unconvinced by the name: I wondered what a `lens` is!

Perhaps try with something like `extents` or `lengths`?

honeycomb-core/src/cmap/builder/grid.rs
```diff
@@ -379,3 +453,286 @@ fn generate_tris_beta_values(n_x: usize, n_y: usize) -> impl Iterator<Item = [Da
...
 }
+
+// ------ 3D
+
+#[allow(unused)]
+/// Internal grid-building routine
+#[allow(clippy::too_many_lines)]
+pub fn build_3d_grid<T: CoordsFloat>(
+    origin: Vertex3<T>,
+    ns: [usize; 3],
```
cedricchevalier19, line 464: Can you have a more explicit name?

`ns` means a lot of stuff in computer science: namespace, ...

Belows you have `n_darts`, so `n_cells_by_axis` or something like that will be more readable.

honeycomb-core/src/cmap/builder/grid.rs
```diff
@@ -379,3 +453,286 @@ fn generate_tris_beta_values(n_x: usize, n_y: usize) -> impl Iterator<Item = [Da
...
+    map
+}
+
+//
+//    y+
+// z+ |
+// \  |
+//  \ |
+//   \|
+//    +------x+
```
cedricchevalier19, line 505: Is this direct? To me, z+ seems to go in the wrong direction.

cedricchevalier19 (commented):
Some comestic remarks.

honeycomb-core/src/cmap/builder/grid.rs
imrn99: Yeah, I made a msitake. I changed the direction of z and updated the coordinates values to make face normals outside-facing.

honeycomb-core/src/cmap/builder/grid.rs
imrn99: renamed to `n_cells_per_axis`

honeycomb-core/src/cmap/builder/grid.rs
imrn99: renamed to `lengths`

honeycomb-core/src/cmap/builder/grid.rs
imrn99: I reversed the changes for now, and commented out the `parse_3d` function. I will deal with this issue in the PR which adds `CMap3` building to the builder struct

honeycomb-core/src/cmap/builder/grid.rs
imrn99: Using dimension as a const generic requires propagating it to `CMapBuilder`. I reversed this for now, and will deal with it when adding the `CMap3` building method the API.

PR #266, opened 2025-01-16
refactor(core): remove dead code from `AttributeStorageManager`

imrn99:
### Description

**Scope**:  core

**Type of change**: refactor

**Content description**:

- make `AttributeStorageManager` only visible in the crate
- remove all of its `force_` variants
- comment out the rest of the unused methods

### Additional information

- [x] Breaking change: technically breaking since I remove `AttributeStorageManager` from the public API. It shouldn't be used as a standalone anyway

### DIFF
--- a/honeycomb-core/src/attributes/manager.rs
+++ b/honeycomb-core/src/attributes/manager.rs
@@ -35,25 +35,6 @@ macro_rules! get_storage {
     };
 }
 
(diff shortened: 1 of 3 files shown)


imrn99:
the dead code represents less of the untested code than I thought. I'll extend tests in another PR

cedricchevalier19 (approved):

PR #269, opened 2025-01-21
feat(benches): add two new binaries to benchmark scaling

imrn99:
### Description

**Scope**:  benches

**Type of change**: feat

**Content description**:
- add the `grid-random-split` binary. it generates a fixed size unit grid with each square cell having a diagonal split, the direction being random. the map is saved as a legacy VTK binary file
- add the `cut-edges` binary. it loads a a map from a VTK file, and divide its cells iteratively until a target edge size is reached. the input mesh is assumed to be a triangle mesh, we guarantee that each intermediate step and final result are as well

### DIFF
--- a/benches/src/cut_edges.rs
+++ b/benches/src/cut_edges.rs
@@ -1,0 +1,103 @@
+use std::time::Instant;
+
+use honeycomb::prelude::{
+    splits::split_edge_noalloc, CMap2, CMapBuilder, CoordsFloat, DartIdType, EdgeIdType,
+};
+
+use honeycomb_core::stm::atomically;
+use rayon::prelude::*;
+
+const INPUT_MAP: &str = "grid_split.vtk";
+const TARGET_LENGTH: f64 = 0.1;
+
+fn fetch_edges_to_process<'a, 'b, T: CoordsFloat>(
+    map: &'a CMap2<T>,
+    length: &'b T,
+) -> impl Iterator<Item = EdgeIdType> + 'a
+where
+    'b: 'a,
+{
+    map.iter_edges().filter(|&e| {
+        let (vid1, vid2) = (
+            map.vertex_id(e as DartIdType),
+            map.vertex_id(map.beta::<1>(e as DartIdType)),
+        );
+        let (v1, v2) = (
+            map.force_read_vertex(vid1).unwrap(),
+            map.force_read_vertex(vid2).unwrap(),
+        );
+
+        (v2 - v1).norm() > *length
+    })
+}
+
+fn main() {
+    let mut instant = Instant::now();
+    // load map from file
+    let mut map: CMap2<f64> = CMapBuilder::from(INPUT_MAP).build().unwrap();
+    println!("map loaded in {}ms", instant.elapsed().as_millis());
+
+    instant = Instant::now();
+    // compute first batch
+    let mut edges: Vec<EdgeIdType> = fetch_edges_to_process(&map, &TARGET_LENGTH).collect();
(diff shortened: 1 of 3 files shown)


honeycomb-core/src/cmap/dim2/basic_ops.rs
```diff
@@ -35,6 +35,28 @@ impl<T: CoordsFloat> CMap2<T> {
         self.unused_darts.iter().filter(|v| v.read_atomic()).count()
     }
 
+    /// Return whether a given dart is unused or not.
+    #[must_use = "unused return value"]
+    pub fn is_unused(&self, d: DartIdType) -> bool {
```
cedricchevalier19, line 40: Is it necessary to have it in a public-facing interface?

benches/src/cut_edges.rs
```diff
@@ -0,0 +1,368 @@
...
+        instant = Instant::now();
+        assert!(
+            map.iter_faces()
+                .all(|f| { Orbit2::new(&map, OrbitPolicy::Face, f as DartIdType).count() == 3 }),
+            "Input mesh isn't a triangle mesh"
+        );
+        println!("topology checked in {}ms", instant.elapsed().as_millis());
+    }
+
+    let backend = if std::env::var("BACKEND").is_ok_and(|s| &s == "rayon") {
```
cedricchevalier19, line 37: Command line arguments should be a more robust approach.

Think of printing the selected backend to confirm how the results were obtained.

cedricchevalier19 (approved):
Some random thoughts.

benches/src/cut_edges.rs
imrn99: I'm thinking of adding `clap` as a dependency and merge all benches in a single binary. then we could have something like:

```sh
hc-bench <COMMON OPTIONS> <COMMAND> <ARGS>
```

with `COMMAND` being the benchmark to run. 

For printing, I had it initially, but I was thinking about adding a csv output. if we want both, I'll need to write directly to a file instead of printing the csv data.

imrn99:
I updated prints and CLI. an execution now looks like this:

```
$ taskset -c 2,4,6,8 ./target/profiling/cut-edges ./grid_split.vtk
Run information
|-> input      : ./grid_split.vtk (hash: 0x216a4a469f2c1f58)
|-> backend    : StdThreads with 4 thread(s)
|-> target size: 0.1
|-> init time  : 13ms
 Step | n_edge_total | n_edge_to_process | t_compute_batch(s) | t_process_batch(s)
    0 |          800 |               800 |        1.050551e-2 |        1.735383e-2
    1 |         3136 |              3136 |        1.783228e-2 |        4.814345e-2
    2 |        12416 |             12416 |        7.197218e-2 |        1.911362e-1
    3 |        49408 |             49408 |        3.241786e-1 |        7.432630e-1
    4 |       197120 |             54368 |         1.107940e0 |        8.808021e-1
    5 |       360224 |             41911 |         2.103729e0 |        6.927806e-1
    6 |       485957 |             21697 |         2.817646e0 |        3.462518e-1
    7 |       551048 |             11062 |         3.222523e0 |        1.688858e-1
    8 |       584234 |              2755 |         3.346779e0 |        4.143390e-2
    9 |       592499 |               503 |         3.413100e0 |        6.631670e-3
   10 |       594008 |                61 |         3.440633e0 |        9.280440e-4
   11 |       594191 |                 0 |         3.422463e0 |         0.000000e0
```

the binary reads cl arguments such that:

```sh
cut-edges <INPUT_FILE> <BACKEND> <TARGET_SIZE>
```

- `BACKEND` can be "rayon" or "chunks", anything else will default to std threads
- the number of threads used is determined using `std::thread::available_parallelism`

PR #270, opened 2025-01-21
chore: add basic Dockerfile

imrn99:
### Description

I used fastiron's file as reference: https://github.com/cea-hpc/fastiron/blob/main/Dockerfile

I adjusted it to:
- automatically clone the repo before build. That way we can restrict space mount to an "out dir"
- clone the repo I use to store input files
- remove the path installation since we're working with many binaries/benchmarks
- use a base with more features for the final image. I want to use the docker in interactive mode, so having a working shell is nice

TODO: test it


Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM rust:1.84 as builder
+
+WORKDIR /builder
+
+# Install git for cloning
+RUN apt-get update && apt-get install -y git
+
+# Clone repositories (honeycomb + inputs)
+RUN git clone https://github.com/LIHPC-Computational-Geometry/honeycomb .
```
cedricchevalier19, line 9: I have mixed feelings about this. If you want to create a container to bench the current branch you are one, it will not do.
Plus, it will not work if you are not connected to github .

I think accessing or copying the current repository into the builder makes more sense. You can use "COPY" or "RUN --mount" (see https://docs.docker.com/reference/dockerfile/#run---mount).

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN git clone https://github.com/imrn99/meshing-samples
```
cedricchevalier19, line 10: For this one, you can use `ADD` to download and decompress the tarball from github. This way git is not required anymore.

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN cargo build --bins --release
+RUN cargo build --bins --profile profiling
+
+# Use Ubuntu as the runtime image
+FROM ubuntu:22.04
+
+# Install performance tools
+RUN apt-get update && apt-get install -y \
+    linux-tools-generic \
+    heaptrack
```
cedricchevalier19, line 23: Does not having the source code available a problem here?

cedricchevalier19 (commented):
It is not totally clear to me what is the goal of this container.

If it is to provide a functional image for the end-user, the location is ok, if it is more for development and benchmarking, I think putting the file elsewhere or renaming it to `benchmark.dockerfile` or something like that is better.

Dockerfile
imrn99: it seems like ADD timeouts on git links from my trials

Dockerfile
imrn99: I switched to binaries install due to issues with generated images

Dockerfile
imrn99: see global comment

imrn99:
I updated the file to be closer to how fastiron's one worked. I think you're right, it will be easier to generate some binaries + mount the repo for the rest of the files.

the perf install didn't work though, I think there is a version incompatibility. If that's ok I will fix it later, when we need actual profiling instead of exec times.

cedricchevalier19 (approved):
I still think that a more precise name like `benchmarks.dockerfile` makes more sense.

imrn99:
> I still think that a more precise name like `benchmarks.dockerfile` makes more sense.

oh ok, I thought this point was related to the original content. I renamed the file

PR #272, opened 2025-01-28
fix(core): replace `while` with marks in 2D ID computations

imrn99:
### Description

**Scope**:  core

**Type of change**: fix 

**Content description**:

the condition for the while loop may never be reached in transactional contexts due to data edits that are not detected until commit time. by reverting to the usual routines used by orbits/3D, we can avoid this issue and guarantee to reach commit time.

### DIFF
--- a/honeycomb-core/src/cmap/dim2/basic_ops.rs
+++ b/honeycomb-core/src/cmap/dim2/basic_ops.rs
@@ -9,6 +9,8 @@
 
 // ------ IMPORTS
 
+use std::collections::{HashSet, VecDeque};
+
 use crate::prelude::{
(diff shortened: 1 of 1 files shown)


cedricchevalier19 (approved):
I think it will be great to list in the documentation what properties are ensured with the transactions.

imrn99:
> I think it will be great to list in the documentation what properties are ensured with the transactions.

I think I will add this when the changes from #274 and #273 are made since they are necessary to ensure the correctness we aim for.

It may also be important to mention that some aspects of transactions are implementation dependent (e.g. some STM impls detect operation on inconsistent data during the transaction while ours detects it at commit time)

PR #273, opened 2025-01-28
refactor!: rework basic operation errors and checks

imrn99:
### Description

**Scope**:  good luck

**Type of change**: refactor

**Content description**:

core crate:
- change the `fast-stm` revision used to have access to the `try_or_coerce!` convenience macro
- replace `CMapError` with the following errors:
	- `LinkError`
	- `AttributeError`
	- `SewError`
- change `AttributeUpdate` associated functions return types to allow errors in each (under the form of an `AttributeError`), update `Vertex2` and `Vertex3` impls
- change the default implementation behavior of `AttributeUpdate::merge_incomplete` to fail
- change transactional methods return types  from `CMapResult` to `TransactionalClosureResult`
- replace inner invariant assertions (e.g. dart is 1-free before a 1-link op) with conditional calls to `abort`
- `force_` variants of link and sew ops now return a `Result` to accommodate the replacement of assertions

kernels crate:
- update `SplitEdgeError`
- update `splits` routines; I did the bare minimum since I want to work on generalized cell insertion at some point
- update `grisubal`

others:
- fix warnings in other crates
- fix all tests

### Additional information

- [x] Breaking change

auto coercion isn't implemented from our errors (e.g. `LinkError`) to `TransactionError::Abort(E)` in order to force user code to acknowledge possible failure points, and think about retry / cancel policies w.r.t. deadlocks / infinite wait scenarios

I haven't modified the prelude of the core crate beacuse I intend to delete it while I sort all other imports.

### Necessary follow-up

- unify behavior between regular and `force` variants (== turn `force` variants into simple wrappers)
    -  remove unused methods from trait impl after this
- remove attribute doctests in favor of a UG page, like the one existing for STM
- add error explanation in the doc
- add properties ensured by transactions in the doc

### DIFF
--- a/honeycomb-kernels/src/splits/edge_multiple.rs
+++ b/honeycomb-kernels/src/splits/edge_multiple.rs
@@ -5,7 +5,10 @@
 use crate::splits::SplitEdgeError;
 use honeycomb_core::cmap::{CMap2, DartIdType, EdgeIdType, NULL_DART_ID};
 use honeycomb_core::geometry::CoordsFloat;
-use honeycomb_core::stm::{atomically, Transaction};
+use honeycomb_core::stm::{
+    abort, atomically_with_err, try_or_coerce, Transaction, TransactionClosureResult,
+};
+
 // ------ CONTENT
 
 #[allow(clippy::missing_errors_doc)]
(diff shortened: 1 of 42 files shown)


honeycomb-kernels/src/splits/edge_multiple.rs
```diff
@@ -106,25 +109,15 @@ pub fn splitn_edge<T: CoordsFloat>(
...
-            match e {
-                SplitEdgeError::FailedTransaction(stme) => Err(stme),
-                SplitEdgeError::UndefinedEdge
-                | SplitEdgeError::VertexBound
-                | SplitEdgeError::InvalidDarts(_)
-                | SplitEdgeError::WrongAmountDarts(_, _) => Ok(Err(e)),
-            }
-        } else {
-            Ok(Ok(()))
-        }
```
imrn99, line 127: this is the single best example of what the patched STM API allows us to do

imrn99 (commented):
I think I did not forget anything in the PR body. We can review this together if needed @cedricchevalier19

cedricchevalier19 (approved):
It looks clearer than the previous version, good job!

PR #284, opened 2025-02-11
refactor(core): remove dead code from attribute items

imrn99:
### Description

**Scope**:  core

**Type of change**: refactor

**Content description**: follow up of #277 

Now that the API has converged, the only methods variants we need are transactional. we can wrap them at the highest level since the user is not supposed to directly interact with attribute containers anyway. 

I will do yet another PR to rename the remaining methods from the various attribute traits since code fro the `sew` impls will be affected


imrn99:
a lot of the additions are from update to the test file. these changes are just noise, important parts are in the other files

cedricchevalier19 (approved):
I like this kind of PR!

PR #287, opened 2025-02-14
feat(kernels): implement skewness computing routines

imrn99:
### Description

**Scope**: kernels

**Type of change**: feat

**Content description**: add a module with 2 new function to the kernel crate

- each function takes a map as an input and compute the skewness of its cell
    - the 2-map function, `compute_cell_skewness_2d` computes face skewness
    - the 3-map function, `compute_cell_skewness_3d` computes face or volume skewness depending on the value of a const argument `N: u8`. Volume computation is currently unimplemented as I could not find a reliable definition for unstructured meshes


honeycomb-kernels/src/skewness/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+///
+/// # Return
+///
+/// The values returned are comprised in the `[0.0; 1.0]` range. `0.0` corresponds to ideal
+/// (equilateral) cells while `1.0` corresponds to degenerate cells.
+///
+/// # Panics
+///
+/// This function will panic if a topological vertex has no associated coordinates.
+pub fn compute_cell_skewness_2d<'a, 'b, T: CoordsFloat>(
```
cedricchevalier19, line 26:
```suggestion
pub fn compute_cells_skewness_2d<'a, 'b, T: CoordsFloat>(
```

honeycomb-kernels/src/skewness/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+///
+/// The values returned are comprised in the `[0.0; 1.0]` range. `0.0` corresponds to ideal
+/// (equilateral) cells while `1.0` corresponds to degenerate cells.
+///
+/// # Panics
+///
+/// This function will panic if:
+/// - `I` is not `2` (or `3`, at the moment),
+/// - a topological vertex has no associated coordinates.
+pub fn compute_cell_skewness_3d<'a, 'b, const N: u8, T: CoordsFloat>(
```
cedricchevalier19, line 94:
```suggestion
pub fn compute_cells_skewness_3d<'a, 'b, const N: u8, T: CoordsFloat>(
```

honeycomb-kernels/src/skewness/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+    match N {
+        // compute faces skewness
+        2 => compute_face_skewness_3d(map),
+        // compute volumes skewness
+        3 => todo!(),
+        _ => panic!(),
+    }
+}
+
+fn compute_face_skewness_3d<'a, 'b, T: CoordsFloat>(
```
cedricchevalier19, line 109: Can't this use the `compute_cells_skewness_2d`?

honeycomb-kernels/src/skewness/tests.rs
```diff
@@ -0,0 +1,13 @@
...
+use crate::skewness::compute_cell_skewness_2d;
+
+#[test]
+fn skewtest() {
+    // squares are equiangular
+    let map: CMap2<f32> = CMapBuilder::unit_grid(2).build().unwrap();
+    assert!(compute_cell_skewness_2d(&map).all(|s| s == 0.0));
+    // triangles aren't; their angles are 90, 45, 45
+    let map: CMap2<f32> = CMapBuilder::unit_triangles(2).build().unwrap();
+    assert!(compute_cell_skewness_2d(&map).all(|s| s == 0.25));
```
cedricchevalier19, line 12:
```suggestion
    assert!(compute_cells_skewness_2d(&map).all(|s| s == 0.25));
```

cedricchevalier19 (commented):
Is it possible to have the same name to compute cell skewness depending on the dimension of the cell?

Also, what about the capability to compute this skewness only for one cell?

I wonder if this iterator way of doing things is such a great idea; for example, we might want to compute this only on a part of the `CMap`, which we could, as a user, filter in advance.

imrn99:
> Is it possible to have the same name to compute cell skewness depending on the dimension of the cell?

I renamed the functions to `compute_face_skewness_Nd` for the moment. I don't plan to implement volume skewness atm (not even sure about the definition?), so I think we can deal with that later.

> Also, what about the capability to compute this skewness only for one cell?
> 
> I wonder if this iterator way of doing things is such a great idea; for example, we might want to compute this only on a part of the `CMap`, which we could, as a user, filter in advance.

You are right, we can define an iterator on the user-side as needed. I rewrote the original functions to only compute the skewness of a single face.

PR #299, opened 2025-03-12
feat(core): implement transactional orbits

imrn99:
### Description

**Scope**: core

**Type of change**: feat

**Content description**:
- add a custom unfold iterator `TryFromFn` to use `?` inside the definition closure
- implement transactional orbit computations. 

Usage is slightly weird since chaining operators isn't be possible due to the closure system. The only two solutions I see for this are:
- remove lazy computations, and check whether anything is invalid before returning an iterator over values (effectively moving the `Result` out of the iterator)
- provide a functions that does what's described above, to still allow using the lazily evaluated iterator (in a for loop for example, where results can be `?`)

### Necessary follow-up

use this content in #298 to expose transactions in the API

### DIFF
--- a/honeycomb-core/src/cmap/dim2/orbits.rs
+++ b/honeycomb-core/src/cmap/dim2/orbits.rs
@@ -33,6 +33,7 @@ impl<T: CoordsFloat> CMap2<T> {
     /// and a `HashSet`. There is a possibility to use static thread-local instances to avoid
     /// ephemeral allocations, but [it would require a guard mechanism][PR].
     ///
+    /// [WIKIBFS]: https://en.wikipedia.org/wiki/Breadth-first_search
     /// [PR]: https://github.com/LIHPC-Computational-Geometry/honeycomb/pull/293
     #[allow(clippy::needless_for_each)]
     #[rustfmt::skip]
@@ -118,6 +119,7 @@ impl<T: CoordsFloat> CMap2<T> {
     }
 
     /// Generic orbit transactional implementation.
+    #[allow(clippy::needless_for_each)]
     pub fn orbit_transac(
         &self,
         t: &mut Transaction,
--- a/honeycomb-core/src/cmap/dim2/orbits.rs
+++ b/honeycomb-core/src/cmap/dim2/orbits.rs
@@ -7,6 +7,7 @@ use std::collections::{HashSet, VecDeque};
 
 use crate::cmap::{CMap2, DartIdType, NULL_DART_ID, OrbitPolicy};
 use crate::geometry::CoordsFloat;
+use crate::stm::{StmClosureResult, StmError, Transaction};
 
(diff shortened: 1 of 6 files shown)


honeycomb-core/src/cmap/dim2/orbits.rs
```diff
@@ -116,11 +118,93 @@ impl<T: CoordsFloat> CMap2<T> {
...
+                match opolicy {
+                    OrbitPolicy::Vertex => {
+                        let b2 = self.beta_transac::<2>(t, d)?;
+                        let b0 = self.beta_transac::<0>(t, d)?;
+                        [
+                            self.beta_transac::<1>(t, b2)?,
+                            self.beta_transac::<2>(t, b0)?,
+                        ]
+                        .into_iter()
+                        .for_each(|im| {
```
cedricchevalier19, line 147: Not convinced that it is the simplest way to write it, there are only two iterations, no?

honeycomb-core/src/cmap/dim2/orbits.rs
```diff
@@ -116,11 +118,93 @@ impl<T: CoordsFloat> CMap2<T> {
...
+                            if marked.insert(im) {
+                                // if true, we did not see this dart yet
+                                // i.e. we need to visit it later
+                                pending.push_back(im);
```
cedricchevalier19, line 151: Can these lines grouped in a function or lambda?

honeycomb-core/src/cmap/dim2/orbits.rs
```diff
@@ -116,11 +118,93 @@ impl<T: CoordsFloat> CMap2<T> {
...
+                    OrbitPolicy::Face => {
+                        [self.beta_transac::<1>(t, d)?, self.beta_transac::<0>(t, d)?]
+                            .into_iter()
+                            .for_each(|im| {
+                                if marked.insert(im) {
+                                    // if true, we did not see this dart yet
+                                    // i.e. we need to visit it later
+                                    pending.push_back(im);
+                                }
+                            });
```
cedricchevalier19, line 177: Same as for vertices

cedricchevalier19 (commented):
Looks like you can factorize further.

honeycomb-core/src/cmap/dim2/orbits.rs
imrn99: I switched to regular statements for 2D. 3D keeps the array + for_each style since it consistently has more elements

honeycomb-core/src/cmap/dim2/orbits.rs
imrn99: I used a lambda to be able to pass the closure directly to for_each when using an iterator

cedricchevalier19 (approved):

PR #304, opened 2025-03-19
refactor!(render): switch to gizmos-based rendering

imrn99:
### Description

**Scope**: render

**Type of change**: refactor

**Content description**:

- remove the capture system, only a single map is rendered now
- remove all mesh-based rendering, switch to gizmos instead
	- unlike meshes, gizmos aren't persistent object and needs to be rendered continuously with a system. this significantly simplifies the parameterization of the render
- remove the picking features for now (I haven't found a simple way to implement it yet)
- remove the `App` object and replace it with a single function
- remove everything related to beta rendering (which wasn't implemented anyway)
- simplify the shrinking system / normal storage in prevision of the 3D implementation
- **remove the dock, settings menu is now opened with tab**

#### Before

Screenshot_20250319_150511

#### After

Screenshot_20250319_150623

### Additional information

- [x] Breaking change


imrn99:
This should make the tool usable for doc / artifact illustrations

PR #311, opened 2025-03-28
feat(core): add new dart allocation interfaces

imrn99:
### Description

should be rebased

**Scope**: core

**Type of change**: feat

**Content description**: see commits

### Necessary follow-up

There are a few different things possible for improvements before integrating some form of [CRDT](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type):

- remove old interfaces
- make block size a parameter of `CMap`s and only allow aligned block allocations
- build the dart blocks during the allocation and add them to a transactional queue that the threads can pop from


honeycomb-core/src/cmap/components/unused.rs
```diff
@@ -22,6 +22,11 @@ impl UnusedDarts {
         self.0.extend((0..len).map(|_| TVar::new(false)));
     }
 
+    /// Extend internal storage capacity using the passed value for new items
+    pub fn extend_from_val(&mut self, len: usize, val: bool) {
```
cedricchevalier19, line 26:
```suggestion
    pub fn extend_with(&mut self, len: usize, val: bool) {
```
Not sure about the semantic you want to follow, but this looks closer to what people are doing externally.

imrn99: yeah, I have no strong opinion on the naming, doing what others do is probably best

I'll need to properly update everything in a single commit if I want to rebase this tho

imrn99:
> can we discuss it more this week?

sure, I also started to look into CRDTs so maybe this can tie things in.

imrn99:
closing in favor of #344

PR #312, opened 2025-03-31
feat(kernels): implement geometrical anchoring attributes

imrn99:
### Description

**Scope**: kernels

**Type of change**: feat

**Content description**:
- add three new enums: `VertexAnchor`, `EdgeAnchor`, and `FaceAnchor`
- implement attribute traits for these

### Necessary follow-up

- ~~add tests~~ done
- set these at the end of the `grisubal` kernel
- use these to implement predicates controlling the `collapse_edge` routine

### DIFF
--- a/honeycomb-kernels/src/remeshing/anchoring.rs
+++ b/honeycomb-kernels/src/remeshing/anchoring.rs
@@ -15,7 +15,7 @@ use honeycomb_core::{
 /// merge two anchors. The merge-ability of two anchors also depends on their intersection; we
 /// expect this to be handled outside of the merge functor, as doing it inside would require leaking
 /// map data into the trait's methods.
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub enum VertexAnchor {
     /// Vertex is linked to a node.
     Node(usize),
@@ -108,7 +108,7 @@ impl AttributeUpdate for VertexAnchor {
 /// merge two anchors. The merge-ability of two anchors also depends on their intersection; we
 /// expect this to be handled outside of the merge functor, as doing it inside would require leaking
 /// map data into the trait's methods.
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub enum EdgeAnchor {
     /// Vertex is linked to a curve.
     Curve(usize),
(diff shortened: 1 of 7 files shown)


honeycomb-kernels/src/remeshing/anchoring.rs
```diff
@@ -0,0 +1,213 @@
...
+///
+/// This enum is used as an attribute to link mesh vertices to entities of the represented geometry.
+///
+/// The `AttributeUpdate` implementation is used to enforce the dimensional conditions required to
+/// merge two anchors. The merge-ability of two anchors also depends on their intersection; we
+/// expect this to be handled outside of the merge functor, as doing it inside would require leaking
+/// map data into the trait's methods.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum VertexAnchor {
+    /// Vertex is linked to a node.
```
cedricchevalier19, line 20: Why everything is `usize`?

cedricchevalier19 (approved):
Two questions:
- is having an additional `Free` state useful?
- You keep the id of the target, is it easy to manage in term of life-time if you modify the targets?

honeycomb-kernels/src/remeshing/anchoring.rs
imrn99: since geometry and mesh are different things, I don't think these entities should be indexed using `VertexIdType`, etc. 

I could change it to `u32` tho, if that's what you meant.

imrn99:
> is having an additional `Free` state useful?

From the discussion I had with @franck-ledoux, I don't think it is. The only use case I can see for this is to use it as a default (e.g. during classification), but having a default could be detrimental further in the remeshing pipeline (e.g. by allowing an algorithm to continue operating while some classification data was lost) 

> You keep the id of the target, is it easy to manage in term of life-time if you modify the targets?

The advantage of IDs is that they do not require a third-party structure to store information after the initial classification. The disadvantage is that we have to be careful about modifying targets if we do something beyond simple (un)sews.

While integrating a reference in attributes could be a way to handle more complex geometry changes, I personally don't like the idea of adding handles towards external data in attribute values; that's also why I'm not handling one of the connectivity criteria in the merge functor, even if I could by adding an `Arc` to the map and duplicate the ID of the dart holding the value. 

Some of my notes go over this choice, I can share the relevant parts.

PR #322, opened 2025-04-15
feat: add a pre-commit hook config file

imrn99:
### Description

**Scope**: repository

**Type of change**: feat

**Content description**: see commits

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,35 @@
+repos:
+  - repo: local
+    hooks:
+      # Check format
+      # Only a check (vs actually formatting) to avoid mistakes
+      - id: cargo-fmt
+        name: check rust code formatting
+        entry: cargo-fmt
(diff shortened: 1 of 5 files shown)


cedricchevalier19 (approved):
Not sure if it should be added, but `act` (https://github.com/nektos/act) allows to run github actions locally in containers.

The pre-commit hook is easier to use, so act is not a replacement.

imrn99:
> Not sure if it should be added, but `act` ([nektos/act](https://github.com/nektos/act)) allows to run github actions locally in containers.
> 
> The pre-commit hook is easier to use, so act is not a replacement.

Yes, `act` and the pre-commit hook don't serve the same purpose. We can add act later, but the idea of the hook is to **quickly** check what the CI goes over, so that there are less surprises when opening a PR. If we had a smaller CI I would have opted for `act` but I think the hook is more adapted here.

PR #326, opened 2025-04-18
Add remesh to hcbench

imrn99:
### Description

**Scope**: benches

**Type of change**: feat

**Content description**:
- add `remesh` subcomand to `hc-bench` with its possibl arguments
- add boundary edge case in the `collapse` op implementation
- fix other `collapse` issues
- add anchor handling to all `remesh` operations
- add error variant for non swappable edges
- add tests over anchor changes of remesh ops
- check for 0-area in `is_orbit_orientation_consistent`

PR #330, opened 2025-05-06
implement the 2D remeshing pipeline

imrn99:
### Description

**Scope**:  kernels / benches

**Type of change**: feat

**Content description**: title + commits

cleanup of #326 

### Necessary follow-up

- add parallelization and implement the backend selection
- add documentation of the pipeline in the user guide

### DIFF
--- a/benches/src/remesh.rs
+++ b/benches/src/remesh.rs
@@ -239,12 +239,13 @@ pub fn bench_remesh<T: CoordsFloat>(args: RemeshArgs) -> CMap2<T> {
             let (l, r) = (e as DartIdType, map.beta::<1>(e as DartIdType));
             let diff = atomically(|t| compute_diff_to_target(t, &map, l, r, args.target_length));
             if diff.abs() < args.target_tolerance {
+                // edge is within target length tolerance; skip the cut/process phase
                 continue;
             }
             let e = map.edge_id(e);
             // process
             if diff.is_sign_positive() {
-                // edge is 20+% longer than target length => cut
+                // edge is longer than target length => cut
                 if map.is_i_free::<2>(e as DartIdType) {
                     let nd = map.add_free_darts(3);
                     let nds: [DartIdType; 3] = std::array::from_fn(|i| nd + i as DartIdType);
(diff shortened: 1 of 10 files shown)


benches/src/remesh.rs
```diff
@@ -0,0 +1,425 @@
...
+            // filter out
+            let (l, r) = (e as DartIdType, map.beta::<1>(e as DartIdType));
+            let diff = atomically(|t| compute_diff_to_target(t, &map, l, r, args.target_length));
+            if diff.abs() < args.target_tolerance {
+                continue;
+            }
+            let e = map.edge_id(e);
+            // process
+            if diff.is_sign_positive() {
+                // edge is 20+% longer than target length => cut
```
cedricchevalier19, line 247: Where does `20%` come from? Where does the check happen?

cedricchevalier19 (approved):
I am eager to see the performance results.

benches/src/remesh.rs
imrn99: 20% comes from a CLI argument which default to`0.2`. the check is done above (L240-243)

I'll change the comment tho, since it's not necessarily 20%

benches/src/remesh.rs
```diff
@@ -0,0 +1,425 @@
...
+) -> StmClosureResult<f64> {
+    let (vid1, vid2) = (map.vertex_id_transac(t, l)?, map.vertex_id_transac(t, r)?);
+    let (v1, v2) =
+        if let (Some(v1), Some(v2)) = (map.read_vertex(t, vid1)?, map.read_vertex(t, vid2)?) {
+            (v1, v2)
+        } else {
+            retry()?
+        };
+    Ok(((v2 - v1).norm().to_f64().unwrap() - target) / target)
+}
```
imrn99, line 408: this computes the relative difference between edge length and target length

benches/src/remesh.rs
```diff
@@ -0,0 +1,425 @@
...
+                // needed as some operations may remove some edges besides the one processed
+                continue;
+            }
+
+            // filter out
+            let (l, r) = (e as DartIdType, map.beta::<1>(e as DartIdType));
+            let diff = atomically(|t| compute_diff_to_target(t, &map, l, r, args.target_length));
+            if diff.abs() < args.target_tolerance {
+                continue;
+            }
```
imrn99, line 243: this is the tolerance check

PR #342, opened 2025-06-02
chore: improve *nix CI

imrn99:
see #341 

this was an attempt at fixing the CI, but I ended up adding interesting stuff to it.

- we now build on both x86_64 and arm based macos runners
- the `fail-fast` option ensures a linux job won't be cancelled after a macos failure, or vice-versa

### DIFF
--- a/.github/workflows/build_nix.yml
+++ b/.github/workflows/build_nix.yml
@@ -11,12 +11,20 @@ env:
 
 jobs:
   build_crates:
-    name: Build crates for ${{ matrix.os }}
+    name: Build crates (${{ matrix.system }})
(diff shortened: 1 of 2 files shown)


imrn99:
@cedricchevalier19 from what I understand, the CI is broken exclusively for arm-based macos runners. I tried using macos 15 and it didn't work either (latest point to macos 14 according to https://github.com/actions/runner-images#available-images)

I noticed that the CI of #334 used macos 14.7.5 while the ones that fail use macos 14.7.6

cedricchevalier19 (approved):

cedricchevalier19:
I think it is better to merge it now and to investigate separately why macos CI is failing.

PR #346, opened 2025-06-12
experiments with jemalloc and rayon

imrn99:
### Description

**Scope**: core / benches

**Type of change**: feat

**Content description**:
- add parallel iterators over i-cells
- add `par-internals` feature to core crate
	- use rayon's `par_extend` and collect from parallel iterators to see if we have some first touch effects
- add `jemalloc` feature to benches
	- see if it affects parallel performance

### Additional information

- [x] New dependency: `tikv-jemallocator`

### Necessary follow-up

run benches with one and/or the other, along with variations on thread binding and hyperthreading.


imrn99:
the CI works again lol

imrn99:
@cedricchevalier19 I think this is worth merging. The features are fairly well contained, and I can make `par-internals` a default feature too in the parallel remeshing PR.

PR #350, opened 2025-07-01
[DO NOT MERGE] 1st parallel remeshing attempt

imrn99:
keeping this PR alive because intermediate commits are interesting


imrn99:
closing until I get back to this alg. I still think it is worth implementing as it cannot result in orientation/cell reversal by design

imrn99:
closing since what I wanted to show is more explicit in #359 (concurrent reservation)

PR #354, opened 2025-07-04
implement parallel remeshing

imrn99:
### Description

See commits

### Necessary follow-up

There are a few performance things to test (most notably, perf of the check of the swap operation).

I think weak scaling will be impossible to properly define, so we'll run strong scaling


imrn99:
I won't delete the other branch yet because I have some interesting intermediate commits for performance analysis

PR #450, opened 2026-03-20
feat(core): return unlinked image dart ID in `unlink` an `unsew` methods

imrn99:
### Description

**Scope**: core

**Type of change**: feat

**Content description**: `unlink` and `unsew` now return the dart ID of the (old) image of `ld`, e.g., for a 2-unsew of dart *d*, the method will return *β<sub>2</sub>(d)* (before the topology update).

This helps avoid redundant reads, and reduce the chance of observing an invalid state by reading manually the value and then unsewing;

It's one of the limit of the STM. STM would detect at commit time that the beta image value changed between the two operations, but the transaction body code would need to account for that inconsistency.

### Necessary follow-up

use it in algorithm (identified use case for `cut-edges`)

### DIFF
--- a/honeycomb-core/src/cmap/components/betas.rs
+++ b/honeycomb-core/src/cmap/components/betas.rs
@@ -227,7 +227,7 @@ impl<const N: usize> BetaFunctions<N> {
         &self,
         t: &mut Transaction,
         ld: DartIdType,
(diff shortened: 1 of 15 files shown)


cedricchevalier19 (approved):
Sorry for the late review.

